PR #8, opened 2017-07-12
Fix bug with Eigen and add missing evaluation of the CoM acceleration

jcarpent:


include/tsid/math/fwd.hpp
```diff
@@ -20,8 +20,13 @@
 
 #include <Eigen/Core>
 
-#define EIGEN_MALLOC_ALLOWED Eigen::internal::set_is_malloc_allowed(true);
-#define EIGEN_MALLOC_NOT_ALLOWED Eigen::internal::set_is_malloc_allowed(false);
+#ifdef EIGEN_RUNTIME_NO_MALLOC
```
proyan: if EIGEN_MALLOC_ALLOWED is conditional on the macro EIGEN_RUNTIME_NO_MALLOC, it should reflect in https://github.com/stack-of-tasks/tsid/blob/master/include/tsid/solvers/solver-HQP-eiquadprog-rt.hxx#L148 as well

proyan: and in 
https://github.com/stack-of-tasks/tsid/blob/master/include/tsid/solvers/solver-HQP-eiquadprog-rt.hxx#L174
https://github.com/stack-of-tasks/tsid/blob/master/include/tsid/solvers/solver-HQP-eiquadprog-rt.hxx#L150

jcarpent: EIGEN_MALLOC_ALLOWED is only for exceptions when EIGEN_RUNTIME_NO_MALLOC is set.
This issue has been solved in Eigen 3.3.x

PR #10, opened 2017-12-07
[joint-posture-task] Fix bug (function isn't supposed to return anyth…

flforget:
…ing)

same name for function and a variable in the function (may create bug/confusion)


olivier-stasse:
:+1:

PR #11, opened 2018-01-17
[upd] Updated tsid with the latest version of pinocchio

cmastalli:
This pull request uses the latest version of pinocchio. Reviewing the code with @flforget we think that we need to include the following line: "se3::framesForwardKinematics(*m_model, data) inside frameJacobianLocal() function". @andreadelprete could you confirm us that this is required?

### DIFF
--- a/src/robots/robot-wrapper.cpp
+++ b/src/robots/robot-wrapper.cpp
@@ -170,14 +170,14 @@ namespace tsid
                                      const Model::JointIndex index,
                                      Data::Matrix6x & J) const
     {
-      return se3::getJacobian<false>(m_model, data, index, J);
+      return se3::getJacobian<se3::WORLD>(m_model, data, index, J);
     }
     
     void RobotWrapper::jacobianLocal(const Data & data,
                                      const Model::JointIndex index,
                                      Data::Matrix6x & J) const
     {
-      return se3::getJacobian<true>(m_model, data, index, J);
+      return se3::getJacobian<se3::LOCAL>(m_model, data, index, J);
     }
     
     SE3 RobotWrapper::framePosition(const Data & data,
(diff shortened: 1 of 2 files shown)


andreadelprete:
You don't need to call it because it's already called in computeAllTerms, which is the method you should call before doing anything else with RobotWrapper.

src/robots/robot-wrapper.cpp
```diff
@@ -249,14 +249,15 @@ namespace tsid
                                           const Model::FrameIndex index,
                                           Data::Matrix6x & J) const
     {
-      return se3::getFrameJacobian<false>(m_model, data, index, J);
+      return se3::getFrameJacobian(m_model, data, index, J);
```
jcarpent: getFrameJacobian only returns the Jacobian of the Frame expressed in the local Frame coordinate system
getFrameJacobian<se3::WORLD> would be equivalent to getJacobian<se3::WORLD> called with frame.parent as joint index information.

cmastalli:
Now we get the Jacobian in the world frame @jcarpent

src/robots/robot-wrapper.cpp
```diff
@@ -249,14 +249,13 @@ namespace tsid
                                           const Model::FrameIndex index,
                                           Data::Matrix6x & J) const
     {
-      return se3::getFrameJacobian(m_model, data, index, J);
+      return se3::getJacobian<se3::WORLD>(m_model, data, index, J);
     }
```
cmastalli: Now we get the Jacobian in the world frame @jcarpent

src/robots/robot-wrapper.cpp
```diff
@@ -249,14 +249,13 @@ namespace tsid
                                           const Model::FrameIndex index,
                                           Data::Matrix6x & J) const
     {
-      return se3::getFrameJacobian(m_model, data, index, J);
+      return se3::getJacobian<se3::WORLD>(m_model, data, index, J);
```
cmastalli: Now we get the Jacobian in the world frame @jcarpent

proyan: You should use m_model.frames[index].parent as the joint index to get the world jacobian of the parent joint.
se3::getJacobian<se3::WORLD>(m_model, data, m_model.frames[index].parent, J);

src/robots/robot-wrapper.cpp
```diff
@@ -249,7 +249,7 @@ namespace tsid
                                           const Model::FrameIndex index,
                                           Data::Matrix6x & J) const
     {
-      return se3::getJacobian<se3::WORLD>(m_model, data, index, J);
+      return se3::getJacobian<se3::WORLD>(m_model, data, m_model.frames[index].parent, J);
```
cmastalli: Now it's using the right joint index

olivier-stasse:
Travis is going fine on 14.04 as well as the unitary test.
Commit accepted

PR #13, opened 2018-01-31
Fix all warnings.

olivier-stasse:
It is a good practice to remove all warnings to:
1/ ease the reading when implementing a new functionnality.
2/ avoid nasty bugs due to the version of the compiler fixing coding mistakes in a different way.

This pull request is passing unitary tests


olivier-stasse:
They are still problems in an unspecified and DEBUG mode.

olivier-stasse:
No warnings on 16.04 LTS they are still some in 14.04 LTS.

PR #22, opened 2018-12-10
Support for contact point and point constraints

jviereck:
Hi,

this PR adds supports for contact points and point constraints (position of a point). In particular this does:

- creates a new `contact-point` class
- creates a `task-point-equality` class
- exposes the new classes in python and rename the current `ContactPythonVisitor` to `Contact6DPythonVisitor`
- fixes a bug in `contact-6d.cpp`, where the constraint was defined too small and thereby the normal force constraint got lost

The `contact-point` and `task-point-equality` work in the world frame. For this, I've added a new `RobotWrapper::frameJacobianLocalWorldOriented`.

So far, I've tested the implementation on a quadruped demo. If the PR looks good overall, I would add a quadruped URDF model and write a basic test using the point contacts.


bindings/python/contacts/contact-point.hpp
```diff
@@ -0,0 +1,158 @@
+#ifndef __tsid_python_contact_6d_hpp__
+#define __tsid_python_contact_6d_hpp__
+
+#include <boost/python.hpp>
+#include <string>
+#include <eigenpy/eigenpy.hpp>
+#include <boost/python/suite/indexing/vector_indexing_suite.hpp>
+
+//
```
jcarpent: This header should appear at the top of the file and I think you should also add the credit of NYU.

jviereck:
hi, 

I've answered some review comments and fixed the code regression (unittest pass again).

I've added a first version of the se3-task-equality with coordinates oriented to the world frame, but I think the current implementation is wrong and I will take a closer look at it.

Question: Instead of creating a second contact-point class, I wonder if we should rather make the existing contact-6d class more flexible. That is, make it possible to provide an arbitrary number of contact points and if there is only a single contact point avoid creating a torque at the contact. This is mostly as 1) the code between contact-6d and contact-point look very much similar and 2) We had the question about providing a 2 point contact formulation (for a 2d walker) yesterday.

andreadelprete:
> Question: Instead of creating a second contact-point class, I wonder if we should rather make the existing contact-6d class more flexible. That is, make it possible to provide an arbitrary number of contact points and if there is only a single contact point avoid creating a torque at the contact. This is mostly as 1) the code between contact-6d and contact-point look very much similar and 2) We had the question about providing a 2 point contact formulation (for a 2d walker) yesterday.

I like the idea. The only reason why I suggested you to create a new class was to make the job simpler.

jviereck:
Hi @andreadelprete , 

I've added a test with the quadruped and added a few last required changes. This is hopefully it. For now, I decided to keep the current code and not generalize the contact-6d.

Could you please have a look and let me know if you want me to change?

src/tasks/task-se3-equality.cpp
```diff
@@ -153,37 +160,45 @@ namespace tsid
...
 
-      // @todo Since Jacobian computation is cheaper in world frame
-      // we could do all computations in world frame
-      m_robot.frameJacobianLocal(data, m_frame_id, m_J);
+        // desired acc in local frame
+        m_a_des = - m_Kp.cwiseProduct(m_p_error.toVector())
+                  - m_Kd.cwiseProduct(m_v_error.toVector())
+                  + m_wMl.actInv(m_a_ref).toVector();
+      } else {
+        m_v_error = m_wMl.act(v_frame) - m_v_ref;  // vel err in local oriented frame.
```
andreadelprete: MINOR The comment should be "vel err in local world-oriented frame"

src/tasks/task-se3-equality.cpp
```diff
@@ -153,37 +160,45 @@ namespace tsid
...
+
+        m_drift = m_wMl.act(m_drift);
+
+        // desired acc in local oriented frame
+        m_a_des = - m_Kp.cwiseProduct(m_p_error.toVector())
```
andreadelprete: Shouldn't you rotate `m_p_error` from local to global before using it here?

src/tasks/task-se3-equality.cpp
```diff
@@ -153,37 +160,45 @@ namespace tsid
...
+                  - m_Kd.cwiseProduct(m_v_error.toVector())
+                  + m_a_ref.toVector();
+
+        m_J = m_wMl.toActionMatrix() * m_J;
```
andreadelprete: I am afraid this might allocate dynamic memory, which is not good for real-time performance...

andreadelprete: You can check if that's really the case using the CMake flag `EIGEN_RUNTIME_NO_MALLOC`

src/tasks/task-se3-equality.cpp
```diff
@@ -196,8 +196,8 @@ namespace tsid
       m_v_ref_vec = m_v_ref.toVector();
       m_v = v_frame.toVector();
 
-      m_constraint.setMatrix(m_mask.transpose() * m_J);
-      m_constraint.setVector(m_mask.transpose() * (m_a_des - m_drift.toVector()));
+      m_constraint.setMatrix(m_mask.asDiagonal() * m_J);
+      m_constraint.setVector(m_mask.asDiagonal() * (m_a_des - m_drift.toVector()));
```
andreadelprete: In line 200 it may be more efficient to use `cwiseProduct` rather than `asDiagonal`

include/tsid/contacts/contact-point.hpp
```diff
@@ -101,6 +101,8 @@ namespace tsid
       void setForceReference(ConstRefVector & f_ref);
       void setRegularizationTaskWeightVector(ConstRefVector & w);
 
+      void useLocalFrame(bool local_frame);
```
andreadelprete: This method should be carefully documented because the implications of switching from local to global frame here are not trivial. For instance, it affects the frame in which the specified contact normal is interpreted.

unittest/tsid-formulation.cpp
```diff
@@ -475,6 +477,165 @@ BOOST_AUTO_TEST_CASE ( test_invdyn_formulation_acc_force )
...
+  std::string contactFrames[] = {
+    "BL_contact", "BR_contact", "FL_contact", "FR_contact"
+  };
+
+  Vector3 contactNormal = Vector3::UnitZ();
+  std::array<ContactPoint*, 4> contacts;
+
+  for (int i = 0; i < 4; i++) {
+    ContactPoint* cp = new ContactPoint("contact_" + contactFrames[i], robot,
+        contactFrames[i], contactNormal, 0.6, 0., 10., w_forceRef);
```
andreadelprete: Here you should use the variables `mu, fMin, fMax` defined above

unittest/tsid-formulation.cpp
```diff
@@ -475,6 +477,165 @@ BOOST_AUTO_TEST_CASE ( test_invdyn_formulation_acc_force )
   cout<<"Desired CoM position: "<<com_ref.transpose()<<endl;
 }
 
+
+BOOST_AUTO_TEST_CASE ( test_contact_point )
+{
+  const double lx = 0.07;
+  const double ly = 0.12;
+  const double lz = 0.105;
```
andreadelprete: `lx, ly, lz` are not used so can be removed

unittest/tsid-formulation.cpp
```diff
@@ -475,6 +477,165 @@ BOOST_AUTO_TEST_CASE ( test_invdyn_formulation_acc_force )
...
+  }
+
+  // Create an HQP solver
+  SolverHQPBase * solver = SolverHQPFactory::createNewSolver(SOLVER_HQP_EIQUADPROG,
+                                                               "solver-eiquadprog");
+  solver->resize(tsid->nVar(), tsid->nEq(), tsid->nIn());
+
+  Vector dv = Vector::Zero(nv);
+  const HQPOutput *sol;
+  for (int i = 0; i < 10; i++) {
```
andreadelprete: For coherence with the other tests you should use `max_it` here instead of the hard-coded value 10

include/tsid/contacts/contact-point.hpp
jviereck: Added comment.

src/contacts/contact-point.cpp
```diff
@@ -45,7 +45,7 @@ ContactPoint::ContactPoint(const std::string & name,
   m_regularizationTaskWeight(regularizationTaskWeight)
 {
   m_weightForceRegTask << 1, 1, 1e-3;
-  m_forceGenMat.resize(3,3);
+  m_forceGenMat.resize(6,3);
```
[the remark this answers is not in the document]
jviereck: Note that the task-se3-equality has size 6, even if the mask is smaller. So the `m_forceGenMat` has the proper size here.

bindings/python/contacts/contact-point.hpp
jviereck: Thanks for pointing this out. I've move the comments to the first line and added "NYU, MPI Tübingen" to the copyright.

jviereck:
Hi @andreadelprete,

thanks for your latest review comments and for approving this PR!

I've pushed a few more changes. Also, it seems I had written some replies to your review comments, but they shows only up as "pending" and I guess were not visible to you yet. Sorry for that.

The latest commits addresses your review comments. A few high level things to point out:

- There were some more bugs in the test. I tried to get the position of a joint instead of a frame. Also, I didn't apply the mask for the task-se3 to restrict the position cordinates of the contact point only and leave the rotation freely. 
- Based on @jcarpent feedback, I've added "NYU, MPI Tübingen" to the copyright. Is this the proper/correct way and okay with you?
- From the implementation side, the only open question for me is if `m_p_error` should get rotated in the `task-se3-equality`. Note my comment at the corresponding line.

Please let me know what you think.

Best,
Julian

src/tasks/task-se3-equality.cpp
andreadelprete: Unfortunately my memory is not very good, so I don't recall this chat. Anyway, the comment in line 177, where `m_p_error` is computed, says "pos err in local frame", so I would trust it. Moreover, at the moment you're using `m_p_error` in the same way regardless of which frame you're doing computations in (see lines 179 and 188). This is clearly wrong, and one of them must be rotated. I think that this is the one that should be rotated, whereas line 179 is fine.

src/contacts/contact-point.cpp
```diff
@@ -112,7 +117,7 @@ void ContactPoint:: updateForceGeneratorMatrix()
   m_forceGenMat.setIdentity();
 }
 
-unsigned int ContactPoint::n_motion() const { return 3; }
+unsigned int ContactPoint::n_motion() const { return 6; }
```
andreadelprete: Shouldn't this be 3?

unittest/tsid-formulation.cpp
```diff
@@ -475,6 +477,165 @@ BOOST_AUTO_TEST_CASE ( test_invdyn_formulation_acc_force )
...
+  std::array<ContactPoint*, 4> contacts;
+
+  for (int i = 0; i < 4; i++) {
+    ContactPoint* cp = new ContactPoint("contact_" + contactFrames[i], robot,
+        contactFrames[i], contactNormal, 0.6, 0., 10., w_forceRef);
+    cp->Kp(kp_contact*Vector::Ones(6));
+    cp->Kd(2.0*cp->Kp().cwiseSqrt());
+    cp->setReference(robot.position(data, robot.model().getJointId(contactFrames[i])));
+    cp->useLocalFrame(false);
+    tsid->addRigidContact(*cp, 1);
```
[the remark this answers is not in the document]
andreadelprete: I don't like it, but I understand that sometimes one needs a quick work around :)
Eventually I'd like to get to the bottom of this issue though.

src/contacts/contact-point.cpp
jviereck: Fixed.

jviereck:
hi @andreadelprete ,

I pushed a new commit which addresses your comments. Note that now the contact-point assumes the mask of the se3-task-equality to be of size 3, which was sort of assumed before as well.

Best,
Julian

src/tasks/task-se3-equality.cpp
```diff
@@ -199,8 +206,18 @@ namespace tsid
       m_v_ref_vec = m_v_ref.toVector();
       m_v = v_frame.toVector();
 
-      m_constraint.setMatrix(m_mask.asDiagonal() * m_J);
-      m_constraint.setVector(m_mask.asDiagonal() * (m_a_des - m_drift.toVector()));
+
+      m_a_des -= m_drift.toVector();
```
andreadelprete: This is a bad idea because now the vector `m_a_des` does not contain the desired acceleration, which the user can read through the method `getDesiredAcceleration`. It'd be better to compute the difference `m_a_des - m_drift` directly in line 217.

src/tasks/task-se3-equality.cpp
```diff
@@ -199,8 +206,18 @@ namespace tsid
...
+
+      int idx = 0;
+      for (int i = 0; i < 6; i++) {
+        if (m_mask(i) != 1.) continue;
+
+        m_constraint.matrix().row(idx) = m_J.row(idx);
+        m_constraint.vector().row(idx) = m_a_des.row(idx);
```
andreadelprete: I think you used the wrong index when accessing `m_J` and `m_a_des`. It should be:
```
m_constraint.matrix().row(idx) = m_J.row(i);
m_constraint.vector().row(idx) = m_a_des.row(i);
```
Probably you couldn't see it in your tests because your mask has all the 1's first, and all the 0's last, and in this case the two implementations are equivalent. If you add a test where the mask is [0, 0, 0, 1, 1, 1]  you should see the difference.

unittest/tsid-formulation.cpp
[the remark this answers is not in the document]
andreadelprete: Hi @jviereck 
after reducing the size of the contact-point constraint to 3, did you check whether it works maintaining the contact motion constraints with priority level 0? Just out of curiosity! :)

andreadelprete:
Hi @jviereck 
I'm testing this PR. Do you have a python script to test the new features using the quadruped URDF model that you have committed?

andreadelprete:
I've seen you've used some C++11 features in the tests. So far the library has been compatible with C++98. We may consider moving to C++11, but I don't know if this is the right place to start this discussion. @jcarpent what is the policy of Pinocchio on this matter?

jcarpent:
Concerning LAAS, the policy is to sick with C++98 for compatibility (HRP-2 notably).
Concerning Pinocchio, it is both C++98, C++11 and C++14 compliant.

andreadelprete:
@jviereck could you please grant me write access to your fork of TSID? So I can push some changes I've just implemented.

jviereck:
@andreadelprete 

> I'm testing this PR. Do you have a python script to test the new features using the quadruped URDF model that you have committed?

Not yet. I can add this if you want.

> @jviereck could you please grant me write access to your fork of TSID? So I can push some changes I've just implemented.

I've added you as a collaborator to the jviereck/tsid repo. Let me know if it doesn't work.

jviereck:
> I'm testing this PR. Do you have a python script to test the new features using the quadruped URDF model that you have committed?

I start working on this now.

jviereck:
@andreadelprete, I pushed a commit with a demo_quadruped.py. The demo shows how to push-up the com_z of the quadruped.

andreadelprete:
Hi @jviereck , thanks for the help. I've added some stuff and pushed everything to master. The main change I've introduced is to let the user specify not only the priority, but also the weight for the motion task (before you were using the same weight for motion and force regularization task, which didn't seem a good idea to me). Moreover, to be coherent with the way weights are specified for Tasks, now the weights associated to a Contact are specified when calling addRigidContact, instead of specifying them when creating the Contact object. The code should be backward compatible though, I've just marked some methods as deprecated and I've added some prints to warn the user to use the new methods.

Next I'm gonna work on the pin-v2 version.

jviereck:
Hi @andreadelprete, thanks for making the final edits, your help along the way for this PR and landing it!

> Moreover, to be coherent with the way weights are specified for Tasks, now the weights associated to a Contact are specified when calling addRigidContact, instead of specifying them when creating the Contact object. The code should be backward compatible though, I've just marked some methods as deprecated and I've added some prints to warn the user to use the new methods.

This sounds good.

andreadelprete:
I've just merged this PR also in the branch pinocchio-v2, in case anybody is interested!

Issue #23, opened 2019-01-24
Not compilable with the new version of Pinocchio


huaijiangzhu:
Hi,

I am getting errors such as: 
error: ‘se3’ does not name a type
It looks like tsid is not adapted to the new namespace in Pinocchio, would you be able to fix this?

huaijiangzhu:
> @huaijiangzhu You just need to add this compilation flag `-DPINOCCHIO_ENABLE_COMPATIBILITY_WITH_VERSION_1`when compiling Pinocchio, nothing more.
> 
> Then someone has to take care of making TSID compatible with the last version of Pinocchio. @huaijiangzhu Can you do it? It is not hard, a simple **find and replace**.

Thanks for the reply! I just replaced the namespace but now I got a new error:
error: no matching function for call to ‘centerOfMass(const Model&, tsid::robots::RobotWrapper::Data&, int, bool)’
       pinocchio::centerOfMass<true,true,true>(m_model, data, false);

Looks like the centerOfMass API is also changed in the new Pinocchio, but I'm not sure how to modify it accordingly. Can you help me with that?

jcarpent:
Yes, sure. For this function, the template parameters have been replaced by a `LEVEL`, see https://github.com/stack-of-tasks/pinocchio/blob/01f211eceda3ac2e5edc8cf101690afb6f3184d3/src/algorithm/center-of-mass.hpp#L101 for further details.
In your case, you want to compute the center of mass position, velocity, and acceleration together, so then LEVEL = 2.

jcarpent:
Because you have to remove the template parameters `<true,true,true>` to obtain:
```
pinocchio::centerOfMass(m_model, data, 2, false);
```

huaijiangzhu:
@jcarpent Thanks! Now the code compiles. But when I try to run the test, I get an import error:

Traceback (most recent call last):
  File "test_Formulation.py", line 2, in <module>
    import tsid
  File "/home/hz/py27/local/lib/python2.7/site-packages/tsid/__init__.py", line 1, in <module>
    import libtsid_pywrap as tsid
ImportError: libtsid.so: cannot open shared object file: No such file or directory
Any idea what can cause this?

jcarpent:
Have you installed TSID? In which directory?
Similarly to Pinocchio, you have to configure your environment variables. You can look at https://stack-of-tasks.github.io/pinocchio/download.html#Install_4

jcarpent:
[...]

PR #25, opened 2019-01-26
Make TSID compatible with Pinocchio 2 

huaijiangzhu:
This PR resolves the issue #23 by replacing the namespace se3 with pinocchio.


jcarpent:
@huaijiangzhu Do you know why your system has changed the right on each file of TSID? Can you revert it?

jviereck:
> What do you mean by "right"?

@huaijiangzhu, if you look at the file changes on GitHub, you see many files have no content change but an ownership/permission change from  100644 → 100755.

huaijiangzhu:
> > What do you mean by "right"?
> 
> @huaijiangzhu, if you look at the file changes on GitHub, you see many files have no content change but an ownership/permission change from 100644 → 100755.

I see. The new commits should fix it!

wxmerkt:
Awesome! Can this be merged before the MEMMO winter school such that people have an easier time installing it?

wxmerkt:
That's a good point! We've also noticed that the Romeo example does not work when run from source.

andreadelprete:
> That's a good point! We've also noticed that the Romeo example does not work when run from source.

Maybe you could open an issue describing the problem(s) you've found? So I can try to help and hopefully everybody can benefit from this.

andreadelprete:
I've checked and I can now confirm that Pinocchio's binaries are still on version 1.3, so merging this PR on master would make TSID incompatible with Pinocchio's binaries. For this reason I think it's better to create a new branch of TSID that is compatible with Pinocchio 2, so that people can choose which one to use depending on their version of Pinocchio.

nim65s:
I can confirm this is the expected schedule :)

andreadelprete:
Too late, I've already merged it in the new branch pinocchio_v2

wxmerkt:
Is there a way to do it with a compile time define guard?

nim65s:
We can put `ADD_REQUIRED_DEPENDENCY("pinocchio < 2.0.0")` in master and `ADD_REQUIRED_DEPENDENCY("pinocchio >= 2.0.0")` in the pinocchio_v2 branch

wxmerkt:
I was more thinking of a way to check for the pinocchio version during compile time and have defines for the API bits that changed. E.g.:

```
#ifdef PINOCCHIO_VERSION_2
using namespace se3 = pinocchio;
#endif
```

or similar - so it can be a single branch instead of two to be maintained.

jcarpent:
Yes, there is such can of mechanism in Pinocchio itself. You just need to add the compilation flag `-DPINOCCHIO_ENABLE_COMPATIBILITY_WITH_VERSION_1`. See #23 for related discussions.

PR #27, opened 2019-01-28
Fix demo_romeo for pinocchio v2

wxmerkt:
Apart from me still getting segmentation fault from `gepetto-viewer-corba`, this fixes the Romeo example


andreadelprete:
Thanks @wxmerkt !

PR #29, opened 2019-01-31
Compatibility with catkinized Pinocchio

wxmerkt:
We are currently working on packaging Pinocchio v2 for catkin, and then releasing it to the ROS buildfarm (including the Python bindings via eigenpy, which will also get released as binaries on ROS). One defect due to limitations in catkin is that the preprocessor directives (for urdfdom, lua, etc.) are not included in the pkg-config generated by catkin - this works for CMake though.

This PR introduces a fail-safe support for both existing Pinocchio and catkinized Pinocchio. It searches _quietly_ for catkin Pinocchio which sets up the `add_definitions`. If not found, the normal procedure is followed.


andreadelprete:
Merged, thanks @wxmerkt !

PR #40, opened 2019-04-10
Convert the exercizes into jupyter notebooks

NoelieRamuzat:
Hello, 

Some phd students worked on the exercises of tsid and we try to convert them into jupyter notebooks.
We changed the robot used, Romeo, by Pyrene and adapted the exercises to use pinocchio v3.
We still have troubles in the control (the robot fly), but @olivier-stasse asks me to put them online to allow people to play with the notebooks (which are more user-friendly).

I have written a [wiki page](https://github.com/NoelieRamuzat/tsid/wiki/Jupyter-Notebooks-Installation) to describe the installation process of jupyter, it may be useful to add it into the tsid wiki if you accept the PR.  

Thank you !


andreadelprete:
Hi @NoelieRamuzat, thanks for this PR. Is there a way you could avoid duplicating the plot_utils.py file and import the one that's already in tsid/exercizes?

NoelieRamuzat:
Hi @andreadelprete, 
Yes I could use the plot_utils.py of tsid/exercizes but I have added a `figsize` argument into the `create_empty_figure()` method, in:
`f, ax = plt.subplots(nRows,nCols, figsize=(10, 10),sharex=sharex);`

 I did this to have a better display in the Jupyter notebooks.
 I can modify the plot_utils.py in tsid/exercizes to take an argument which may specify the size we want, such as:
`create_empty_figure(nRows=1, nCols=1, figsize=(10,10), spinesPos=None,sharex=True)`

Which solution do you prefer ?

andreadelprete:
Yes, please modify the original `create_empty_figure` function.

NoelieRamuzat:
I made the modifications, tell me if something should be changed.

PR #42, opened 2019-04-10
Add TrajSample.pos(SE3) for python bindings.

jviereck:
This allows to use an SE3 object to set the trajectorie's sample position. 

I was using before this python code to convert the 7-dim q vector to the required 12-dim se3-position vector

```
def base_pos_se3_vector(q):
    """ Convert the q base position into an SE3 object. """
    quad = se3.Quaternion(float(q[6]), float(q[3]), float(q[4]), float(q[5]))
    pos = zero(12)
    pos[:3] = q[:3].reshape(-1, 1)
    pos[3:] = quad.matrix().reshape(-1, 1)
    print(quad.matrix())
    # Yields:
    # [[ 1.    0.    0.  ]
    #  [-0.    1.   -0.04]
    #  [-0.    0.04  1.  ]]
    return pos
```

however, this does not yield the desired / expected result. The reshaping of the `quad.matrix` is done row-wise, where the C++ code in `tsid::math::SE3ToVector` seems to do it column-wise. In practice, this caused issues when tracking a desired orientation. In the above example, the python code would return 

```
pos.T: matrix([[ 0.  ,  0.64,  0.22,  1.  ,  0.  ,  0.  , -0.  ,  1.  , -0.04, -0.  ,  0.04,  1.  ]])
```
where propagating the same q into the invdyn and reading the desired position from the se3-task.position yields

```
pos.T: matrix([[ 0.  ,  0.64,  0.22,  1.  ,  0.  ,  0.  , -0.  ,  1.  , 0.04, -0.  ,  -0.04,  1.  ]])
```

Note the flip of sign at the +-0.04.

PS1: This is my new `base_pos_se3_vector` with the more fail-safe SE3 object:

```
def base_pos_se3_vector(q):
    """ Convert the q base position into an SE3 object. """
    quad = se3.Quaternion(float(q[6]), float(q[3]), float(q[4]), float(q[5]))
    return se3.SE3(quad.matrix(), q[:3].reshape(-1, 1))
```

PS2: Is there a way to get the SE3 object from the 7 entry q vector using a single build in method form pinocchio?

### DIFF
--- a/bindings/python/trajectories/trajectory-base.hpp
+++ b/bindings/python/trajectories/trajectory-base.hpp
@@ -23,43 +23,50 @@
 #include <eigenpy/eigenpy.hpp>
 #include <boost/python/suite/indexing/vector_indexing_suite.hpp>
 
+#include <tsid/math/utils.hpp>
 #include "tsid/trajectories/trajectory-base.hpp"
 #include <assert.h>
 namespace tsid
 {
   namespace python
-  {    
+  {
     namespace bp = boost::python;
-    
+    typedef pinocchio::SE3 SE3;
+
     template<typename TrajSample>
     struct TrajectorySamplePythonVisitor
     : public boost::python::def_visitor< TrajectorySamplePythonVisitor<TrajSample> >
     {
-      
-      template<class PyClass>     
+
+      template<class PyClass>
 
       void visit(PyClass& cl) const
       {
         cl
         .def(bp::init<unsigned int>((bp::arg("size")), "Default Constructor with size"))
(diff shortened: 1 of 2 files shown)


jviereck:
On one of my examples I am getting a segfault with this patch - please don't merge yet.

jviereck:
The problem of the segfault was that `TrajectorySE3Constant` did not always initialize the sample. I've added a commit to init the sample always (which can be done in the SE3 case as the dimensionality should not change.

andreadelprete:
Hi @jviereck , thanks for this PR which indeed makes the use of SE3Trajectory more fail safe. To answer your question, unfortunately I don't know any way to get the SE3 object from the 7 entry q vector using a single build in method from pinocchio. Maybe @jcarpent does?

PR #44, opened 2019-04-17
Topic/angular momentum

ggory15:
Recently, I made task function for angular momentum.
The task is exactly same as that in the pinocchio-inv-dyn.
 
More specifically,
1. "task-angular-momentum-equality.cpp" was generated with "TaskBase" class. 
   (Existed tasks such as "com", "jointposture", and "EE" are based on "TaskMotion" class. 
    If you don't like this new task based on "TaskBase", Please tell me. I will modify it. )
2. I made a few of unit-test codes which are related on angular momentum task.
3. I added angular momentum task in the demo_romeo.py and demo_quadprog.py.

ps. [stack-of-tasks with devel branch] The unit-test, tsid-formulation.cpp, was not passed in my computer. I don't know if this error is only a problem for my computer.

### DIFF
--- a/include/tsid/formulations/inverse-dynamics-formulation-base.hpp
+++ b/include/tsid/formulations/inverse-dynamics-formulation-base.hpp
@@ -24,6 +24,7 @@
 #include "tsid/tasks/task-actuation.hpp"
 #include "tsid/tasks/task-motion.hpp"
 #include "tsid/tasks/task-contact-force.hpp"
+#include "tsid/tasks/task-angular-momentum-equality.hpp"
 #include "tsid/contacts/contact-base.hpp"
 #include "tsid/solvers/solver-HQP-base.hpp"
 
@@ -45,6 +46,7 @@ namespace tsid
     typedef math::RefVector RefVector;
     typedef math::ConstRefVector ConstRefVector;
     typedef tasks::TaskMotion TaskMotion;
+    typedef tasks::TaskAMEquality TaskAngularMomentum;
     typedef tasks::TaskContactForce TaskContactForce;
     typedef tasks::TaskActuation TaskActuation;
     typedef tasks::TaskBase TaskBase;
@@ -65,6 +67,11 @@ namespace tsid
     virtual unsigned int nIn() const = 0;
 
     virtual bool addMotionTask(TaskMotion & task,
+                               double weight,
+                               unsigned int priorityLevel,
+                               double transition_duration=0.0) = 0;
+
+    virtual bool addMomentumTask(TaskAngularMomentum & task,
                                double weight,
                                unsigned int priorityLevel,
                                double transition_duration=0.0) = 0;
--- a/include/tsid/tasks/task-angular-momentum-equality.hpp
+++ b/include/tsid/tasks/task-angular-momentum-equality.hpp
@@ -1,0 +1,93 @@
+//
+// Copyright (c) 2017 CNRS
+//
+// This file is part of tsid
+// tsid is free software: you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation, either version
+// 3 of the License, or (at your option) any later version.
+// tsid is distributed in the hope that it will be
+// useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+// of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Lesser Public License for more details. You should have
+// received a copy of the GNU Lesser General Public License along with
+// tsid If not, see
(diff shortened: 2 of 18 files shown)


jviereck:
Hi @ggory15, thanks for providing this PR! This is a feature I was hoping to get supported by TSID as well :)

From your description it sounds like the task is tracking only the angular momentum. Would it be possible to track the linear momentum with the same task as well? If you are only interested to track the angular moment, you could set the grains for the linear momentum to zeros and thereby get an angular momentum task.

src/tasks/task-angular-momentum-equality.cpp
```diff
@@ -0,0 +1,174 @@
...
+                + m_ref.acc;
+
+#ifndef NDEBUG
+      std::cout<<m_name<<" errors: "<<m_L_error.norm()<<" "
+        <<m_dL_error.norm()<<std::endl;
+#endif
+      // Del Prete's quick and dirty way to compute drift
+      // compute momentum Jacobian at next time step assuming zero acc
+      double dt = 1e-3;
+      const Vector & q_next = pinocchio::integrate(m_robot.model(), q, dt*v);
```
andreadelprete: This trick was acceptable in my personal python library, but it's too dirty for TSID. I think you should compute the drift in a proper way. Maybe @jcarpent can suggest you which function of Pinocchio to use to do that.

src/tasks/task-angular-momentum-equality.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    const ConstraintBase & TaskAMEquality::compute(const double ,
+                                                    ConstRefVector q,
+                                                    ConstRefVector v,
+                                                    const Data & data)
+    {
+      // Compute errors
+      // Get momentum jacobian
+      Data data_ccrba = data;
+      pinocchio::ccrba(m_robot.model(), data_ccrba, q, v);
```
andreadelprete: You shouldn't call Pinocchio's function directly, but use the robot wrapper instead.

src/formulations/inverse-dynamics-formulation-acc-force.cpp
```diff
@@ -142,6 +142,29 @@ bool InverseDynamicsFormulationAccForce::addMotionTask(TaskMotion & task,
   return true;
 }
 
+bool InverseDynamicsFormulationAccForce::addMomentumTask(TaskAngularMomentum & task,
```
andreadelprete: It seems to me that the AM task is not different from any other motion task, so I don't see the reason for adding custom methods dedicated to it in `InverseDynamicsFormulationAccForce`. Is there a good reason that I am missing?

include/tsid/tasks/task-angular-momentum-equality.hpp
```diff
@@ -0,0 +1,93 @@
...
+                                     ConstRefVector q,
+                                     ConstRefVector v,
+                                     const Data & data);
+
+      const ConstraintBase & getConstraint() const;
+
+      void setReference(const TrajectorySample & ref);
+      const TrajectorySample & getReference() const;
+
+      const Vector3 & getDesireddMomentum() const;
```
andreadelprete: You wrote `Desiredd` instead of `Desired`

include/tsid/formulations/inverse-dynamics-formulation-base.hpp
```diff
@@ -69,6 +71,11 @@ namespace tsid
                                unsigned int priorityLevel,
                                double transition_duration=0.0) = 0;
 
+    virtual bool addMomentumTask(TaskAngularMomentum & task,
```
andreadelprete: No need for this if you implement the AM task as a MotionTask

andreadelprete:
> Hi @ggory15, thanks for providing this PR! This is a feature I was hoping to get supported by TSID as well :)
> 
> From your description it sounds like the task is tracking only the angular momentum. Would it be possible to track the linear momentum with the same task as well? If you are only interested to track the angular moment, you could set the grains for the linear momentum to zeros and thereby get an angular momentum task.

Setting the gains to zero wouldn't be enough, but you can get what you want by using the mask

ggory15:
I will fix this function with suggestions. (after next monday.)

olivier-stasse:
Can I close this PR ?

stonneau:
@ggory15 Hi Sanghyun, do you think you would have time to update the PR, or should we find a way to do it ourselves ? Thanks a lot !

ggory15:
> @ggory15 Hi Sanghyun, do you think you would have time to update the PR, or should we find a way to do it ourselves ? Thanks a lot !

I am sorry that I totally forget to update it. Because of my ph.d defense, I can't consentrate on this work. Can you close this PR? I will update it later.

PR #46, opened 2019-04-19
[robotwrapper] create constructor from pinocchio::model

proyan:
Hi all,

I think this would be a useful constructor to have, in order to create a robotwrapper from the pinocchio model.
Please let me know if any changes are required.

### DIFF
--- a/src/robots/robot-wrapper.cpp
+++ b/src/robots/robot-wrapper.cpp
@@ -44,6 +44,16 @@ namespace tsid
       m_Md.setZero(m_model.nv);
       m_M.setZero(m_model.nv, m_model.nv);
     }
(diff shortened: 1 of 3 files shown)


src/robots/robot-wrapper.cpp
```diff
@@ -44,6 +44,16 @@ namespace tsid
       m_Md.setZero(m_model.nv);
       m_M.setZero(m_model.nv, m_model.nv);
     }
+    RobotWrapper::RobotWrapper(const pinocchio::Model& m, bool verbose)
```
jviereck: The argument `verbose` is not used here.

proyan: I did. Thanks for finding it.

jviereck:
Thanks @proyan for contributing this change! This is really useful to have :)

PR #55, opened 2019-10-23
Make notebook Python 3 compatible.

jmirabel:


andreadelprete:
Thanks for the work @jmirabel !

PR #69, opened 2020-02-16
Angular momentum equality task

pFernbach:
Hello, I took back the original work from @ggory15 from his previous closed PR (https://github.com/stack-of-tasks/tsid/pull/44) and tried to address all the changes requested. 

The main change that remain to be done is here: 
```C++
      // Del Prete's quick and dirty way to compute drift
      // compute momentum Jacobian at next time step assuming zero acc
      double dt = 1e-3;
      const Vector & q_next = pinocchio::integrate(m_robot.model(), q, dt*v);
      Data data_next = data;
      m_robot.computeAllTerms(data_next, q_next, v);
      const Matrix6x & J_am_next = m_robot.momentumJacobian(data_next);
      m_drift = (J_am_next.bottomRows(3) - J_am.bottomRows(3))* v / dt;
```

If this way of computing the dirft is not acceptable, maybe @jcarpent could give me some pointers to Pinocchio methods to achieve that properly ?

### DIFF
--- a/src/tasks/task-angular-momentum-equality.cpp
+++ b/src/tasks/task-angular-momentum-equality.cpp
@@ -130,9 +130,9 @@ namespace tsid
       // compute momentum Jacobian at next time step assuming zero acc
       double dt = 1e-3;
       const Vector & q_next = pinocchio::integrate(m_robot.model(), q, dt*v);
-      Data data_ccrba = data;
-      pinocchio::ccrba(m_robot.model(), data_ccrba, q_next, v);
-      const Matrix6x & J_am_next = m_robot.momentumJacobian(data_ccrba);
+      Data data_next = data;
+      m_robot.computeAllTerms(data_next, q_next, v);
+      const Matrix6x & J_am_next = m_robot.momentumJacobian(data_next);
       m_drift = (J_am_next.bottomRows(3) - J_am.bottomRows(3))* v / dt;
 
       m_constraint.setMatrix(J_am.bottomRows(3));
--- a/src/tasks/task-angular-momentum-equality.cpp
+++ b/src/tasks/task-angular-momentum-equality.cpp
@@ -109,7 +109,7 @@ namespace tsid
     }
 
     const ConstraintBase & TaskAMEquality::compute(const double ,
-                                                    ConstRefVector q,
+                                                    ConstRefVector ,
                                                     ConstRefVector v,
                                                     const Data & data)
     {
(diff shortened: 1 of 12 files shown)


src/tasks/task-angular-momentum-equality.cpp
```diff
@@ -0,0 +1,145 @@
...
+      m_L_error = m_L - m_ref.vel;
+
+      m_dL_des = - m_Kp.cwiseProduct(m_L_error)
+                + m_ref.acc;
+
+#ifndef NDEBUG
+//      std::cout<<m_name<<" errors: "<<m_L_error.norm()<<" "
+//        <<m_dL_error.norm()<<std::endl;
+#endif
+      // Del Prete's quick and dirty way to compute drift
```
andreadelprete: You can probably use [this function](https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/namespacepinocchio.html#a58aae3c6002a58f93c16b80a81c2e9bb) to compute the rate of change of the (angular) momentum with zero joint accelerations, which is the `drift`. You can use the "quick&dirty" computation in a test to check that what you are computing is indeed correct.

pFernbach: I did it in my latests commits. The difference between the results of the two methods are between 1e-3 and 4e-3. 

Should I add this method to the RobotWrapper class ?

pFernbach:
It looks like travis is configured to use pinocchio 2.1.9 and the method `computeCentroidalMomentumTimeVariation` was introduced in 2.2.0. Should we change the minimal version of Pinocchio dependandy ?

nim65s:
yes :)

src/tasks/task-angular-momentum-equality.cpp
andreadelprete: Yes please, you can add it to RobotWrapper and then you're good to go for me! :)

jcarpent: @pFernbach I think you are not doing the finite differences in the correct way. You should take into account the displacement due to the motion of the frame where the quantity is expressed.

pFernbach: @andreadelprete I pushed the changes

@jcarpent Ok that explain the small differences between this and the `pinocchio::computeCentroidalMomentumTimeVariation` method. Anyway I am now using the method from pinocchio and I removed this finite difference code.

andreadelprete: > @pFernbach I think you are not doing the finite differences in the correct way. You should take into account the displacement due to the motion of the frame where the quantity is expressed.

Agreed. That's why the comment said "quick & dirty"! :)

PR #72, opened 2020-03-09
Release/1.2.3

nim65s:
Same as stack-of-tasks/dynamic-graph#63 : please don't merge :)

Among other changes:
- use cmake-exports from dependencies
- install cmake-exports for this package
- scripts have been moved to tests/python and used as such
- the python module is now `tsid.so`, instead of
```
tsid/
├── __init__.py
└── libtsid_pywrap.so
```


jcarpent:
@nim65s Please do not rename the Python package as tsid.so. It won't be compatible with conda for instance.

nim65s:
ok

nim65s:
But I'd really like to understand that better. Do you have any resource for claim ?

jcarpent:
Not really, but this what I had to fix.

nim65s:
Ok, then let's just keep things ugly and hate conda a bit more :)

jcarpent:
I think the issue was for Windows if I remember well. And you cannot do for instance tsid.__file__

nim65s:
Oops, I forgot that, sorry.

PR #75, opened 2020-03-30
[ROS] Add package.xml

nim65s:
Hi,

This was asked by @olivier-stasse to be able to use this package with catkin build.

@andreadelprete @jcarpent @jviereck  : I put your names in the author field. Please let me know if we have to change anything.


jviereck:
Hi, thanks for putting in as author. I've "only" contributed point contacts to the library. Not sure that is enough to become an author.

What do you think @andreadelprete ?

andreadelprete:
> Hi, thanks for putting in as author. I've "only" contributed point contacts to the library. Not sure that is enough to become an author.
> 
> What do you think @andreadelprete ?

I think @jviereck made an inportant contribution to the project, which deserves to be acknowledged, but so did also @ggory15 (see https://github.com/stack-of-tasks/tsid/graphs/contributors). Maybe we can distinguish between authors and contributors? If so, I'd say @jcarpent and I are authors, while @jviereck and @ggory15 are contributors.

jviereck:
+1 on being a contributor.

nim65s:
I totally agree, but it seems that [ROS' package.xml]( http://wiki.ros.org/catkin/package.xml) doesn't know about contributors.

I suggest to keep @jcarpent & @andreadelprete as authors on the package.xml, and add a section in the readme for authors & contributors.

PR #78, opened 2020-04-03
Joint limits

dalinel:
Hello,
I have added the joint position limits 
It is exactly following [the python code](https://github.com/andreadelprete/pinocchio_inv_dyn/blob/master/python/pinocchio_inv_dyn/acc_bounds_util.py)

I have just changed this in computeAccLimitsFromPosLimits : 
If we exactly reach qmin or qmax I bound the (lower or upper) acceleration with zero to prevent the robot to take the maximum acceleration possible and have a harsh motion

If you want here is a test with Talos : [test](https://github.com/resibots/jointLimitsTsidTest)

### DIFF
--- a/src/tasks/task-joint-posVelAcc-bounds.cpp
+++ b/src/tasks/task-joint-posVelAcc-bounds.cpp
@@ -18,6 +18,13 @@
 #include <tsid/tasks/task-joint-posVelAcc-bounds.hpp>
 #include "tsid/robots/robot-wrapper.hpp"
 
+/** This class has been implemented following :
+* Andrea del Prete. Joint Position and Velocity Bounds in Discrete-Time 
+* Acceleration/Torque Control of Robot Manipulators. IEEE Robotics and Automation
+* Letters, IEEE 2018, 3 (1), pp.281-288.￿10.1109/LRA.2017.2738321￿. hal-01356989v3
+* And
+* https://github.com/andreadelprete/pinocchio_inv_dyn/blob/master/python/pinocchio_inv_dyn/acc_bounds_util.py
+*/
 namespace tsid
 {
   namespace tasks
--- a/src/tasks/task-joint-posVelAcc-bounds.cpp
+++ b/src/tasks/task-joint-posVelAcc-bounds.cpp
@@ -17,6 +17,7 @@
 
 #include <tsid/tasks/task-joint-posVelAcc-bounds.hpp>
 #include "tsid/robots/robot-wrapper.hpp"
+// #include <tsid/utils/stop-watch.hpp>
 
 /** This class has been implemented following :
 * Andrea del Prete. Joint Position and Velocity Bounds in Discrete-Time 
@@ -35,16 +36,17 @@ namespace tsid
 
     TaskJointPosVelAccBounds::TaskJointPosVelAccBounds(const std::string & name,
                                      RobotWrapper & robot,
-                                     double dt):
+                                     double dt,
+                                     bool verbose):
       TaskMotion(name, robot),
       m_constraint(name, robot.na(), robot.nv()),
       m_dt(2*dt),
+      m_verbose(verbose),
       m_nv(robot.nv()),
       m_na(robot.na())
     {
       assert(dt>0.0);
-      eps = 1e-10;
-      resetVectors();
+      m_eps = 1e-10;
       m_qMin=Vector::Constant(m_na,1,1e10);
       m_qMax=Vector::Constant(m_na,1,-1e10);
       m_dqMax=Vector::Constant(m_na,1,1e10);
@@ -53,7 +55,6 @@ namespace tsid
       m_impose_velocity_bounds=false;
       m_impose_viability_bounds = false;
       m_impose_acceleration_bounds = false;
-      m_verbose = true;
 
       //Used in computeAccLimitsFromPosLimits
       m_two_dt_sq = 2.0/(m_dt*m_dt);
@@ -71,7 +72,7 @@ namespace tsid
       m_dt_two_dq  = Vector::Zero(m_na);
       m_two_ddqMax = Vector::Zero(m_na);
       m_dt_ddqMax_dt = Vector::Zero(m_na);
-      m_dq_square = 0.0;
+      m_dq_square = Vector::Zero(m_na);
       m_q_plus_dt_dq = Vector::Zero(m_na);
       m_b_1 = Vector::Zero(m_na);
       m_b_2 = Vector::Zero(m_na);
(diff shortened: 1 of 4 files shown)


src/tasks/task-joint-posVelAcc-bounds.cpp
```diff
@@ -0,0 +1,491 @@
...
+      // std::cout << "rows dim " << dim <<std::endl;
+      // std::cout << "cols dim " <<  m_robot.nv() <<std::endl;
+      // std::cout << m_constraint.vector() << std::endl;
+      // std::cout << m_constraint.vector().size() << std::endl;
+      // std::cout << m_constraint.matrix() << std::endl;
+
+    }
+
+    int TaskJointPosVelAccBounds::dim() const
+    { return m_nv; }
```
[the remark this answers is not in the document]
andreadelprete: Actually I see that the size of the constraint is `na`, so the `dim` method should return `na` instead of `nv`, shouldn't it?

src/tasks/task-joint-posVelAcc-bounds.cpp
```diff
@@ -0,0 +1,491 @@
...
+          {
+            std::cerr << "State of joint "<< i <<" is not viable because q[i]<qMin[i] : "
+            << qa[i] <<"<"<< m_qMin[i] <<std::endl;
+          }
+          if(qa[i] < (m_qMax[i] + eps))
+          {
+            std::cerr << "State of joint "<< i <<" is not viable because qa[i]>m_qMax[i] : "
+            << qa[i] <<">"<< m_qMax[i] <<std::endl;
+          }
+          m_dqMaxViab[i] = std::min( m_dqMax[i],  std::sqrt(std::max(0.0,2*m_ddqMax[i]*(m_qMax[i]-qa[i]))));
```
andreadelprete, line 268: This line is inside the `if(verbose)` branch. Is that done on purpose? 
Moreover indentation is off.

src/tasks/task-joint-posVelAcc-bounds.cpp
```diff
@@ -0,0 +1,491 @@
...
+  }
+
+
+    void TaskJointPosVelAccBounds::computeAccLimitsFromPosLimits(const Vector&q,
+                                                                const Vector& dq,
+                                                                bool verbose)
+  {
+        Vector qa = q.tail(m_na);
+        Vector dqa = dq.tail(m_na);
+        double two_dt_sq   = 2.0/(std::pow(m_dt,2));
```
andreadelprete, line 281: Don't use `std::pow` because it's very slow. Just do `dt*dt`

src/tasks/task-joint-posVelAcc-bounds.cpp
```diff
@@ -0,0 +1,491 @@
...
+        Vector ddqMax_q3 = two_dt_sq*(m_qMax-qa-m_dt*dqa);
+        Vector ddqMin_q3 = two_dt_sq*(m_qMin-qa-m_dt*dqa);
+        Vector ddqMax_q2 = Vector::Zero(m_na);
+        Vector ddqMin_q2 = Vector::Zero(m_na);
```
andreadelprete, line 285: All memory allocation should occur in the initialization phase to make run-time execution as efficient as possible.

src/tasks/task-joint-posVelAcc-bounds.cpp
```diff
@@ -0,0 +1,491 @@
...
+          if(dqa[i]<=0.0)
+          {
+              m_ddqUBPos[i]  = ddqMax_q3[i];
+              if(ddqMin_q3[i] < minus_dq_over_dt[i])
+              {
+                  m_ddqLBPos[i]  = ddqMin_q3[i];
+              }
+              else if(qa[i]!=m_qMin[i])
+              {
+                  ddqMin_q2[i] = std::pow(dqa[i],2)/(2.0*(qa[i]-m_qMin[i]));
```
andreadelprete, line 298: Avoid using std::pow

src/tasks/task-joint-posVelAcc-bounds.cpp
```diff
@@ -0,0 +1,491 @@
...
+          else
+          {
+              m_ddqLBPos[i]  = ddqMin_q3[i];
+              if(ddqMax_q3[i] > minus_dq_over_dt[i])
+              {
+                  m_ddqUBPos[i]  = ddqMax_q3[i];
+              }
+              else if(qa[i]!=m_qMax[i])
+              {
+                  ddqMax_q2[i] = -std::pow(dqa[i],2)/(2*(m_qMax[i]-qa[i]));
```
andreadelprete, line 319: avoid using pow

src/tasks/task-joint-posVelAcc-bounds.cpp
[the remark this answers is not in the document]
dalinel: Yes the bounds needs to be of size na
The accelerations of the floating base are not constrained
The constraint matrix is of size na*nv with the first (nv-na) cols which are zeros 
I will change dim to na

dalinel:
Hello,
Thanks for the review, I will change this and come back to you

dalinel:
Hello,
I have added the changes
I have also tried a full vectorized version but it was slower because of the many if else statements
In the end I have just vectorized some parts

src/tasks/task-joint-posVelAcc-bounds.cpp
```diff
@@ -176,226 +195,182 @@ namespace tsid
...
-        {
-        if(verbose)
+    void TaskJointPosVelAccBounds::computeAccLimitsFromPosLimits(const Vector&qa,
+                                                                  const Vector& dqa,
+                                                                  bool verbose)
+    {
+      m_ddqMax_q3 = m_two_dt_sq*(m_qMax-qa-m_dt*dqa);
+      m_ddqMin_q3 = m_two_dt_sq*(m_qMin-qa-m_dt*dqa);
+      m_ddqMax_q2 = Vector::Zero(m_na);
+      m_ddqMin_q2 = Vector::Zero(m_na);
```
andreadelprete, line 263: Have you verified whether this does not allocate dynamic memory? maybe using the method setZero would be better

src/tasks/task-joint-posVelAcc-bounds.cpp
```diff
@@ -176,226 +195,182 @@ namespace tsid
...
+      m_minus_dq_over_dt = -dqa/m_dt;
+      m_dt_two_dq = 2*m_dt_dq;
+      m_two_ddqMax = 2*m_ddqMax;
+      m_dt_ddqMax_dt = m_ddqMax*m_dt_square;
+      m_dq_square = dqa.dot(dqa);
+      m_q_plus_dt_dq = qa + m_dt_dq;
+      m_b_1 = m_dt_two_dq + m_dt_ddqMax_dt;
+      m_b_2 = m_dt_two_dq - m_dt_ddqMax_dt;
+      m_ddq_1 = Vector::Zero(m_na);
+      m_ddq_2 = Vector::Zero(m_na);
```
andreadelprete, line 327: same here about setZero

src/tasks/task-joint-posVelAcc-bounds.cpp
```diff
@@ -176,226 +195,182 @@ namespace tsid
...
-        double two_a = 2*dt_square;
-        Vector b = dt_two_dq + dt_ddqMax_dt;
-        Vector bl = dt_two_dq - dt_ddqMax_dt;
-        for(int i=0; i<m_na; i++)
+      m_dt_dq = m_dt*dqa;
+      m_minus_dq_over_dt = -dqa/m_dt;
+      m_dt_two_dq = 2*m_dt_dq;
+      m_two_ddqMax = 2*m_ddqMax;
+      m_dt_ddqMax_dt = m_ddqMax*m_dt_square;
+      m_dq_square = dqa.dot(dqa);
```
andreadelprete, line 322: just to check, are you sure you shouldn't use cwiseProduct here instead of dot?

dalinel: Yes that is correct it was cwiseProduct

dalinel:
Hello,
I have done the correctionsn now there is no dynamic memory allocation in runtime
And I have corrected  the cwiseProduct mistake

andreadelprete:
This looks perfect! Thanks a lot for your contribution!

PR #83, opened 2020-05-04
Cheatsheet

ABonnefoy:


nim65s:
This is really nice, thanks.

It would be cool to have it available through doxygen, to avoid the emptiness of https://gepettoweb.laas.fr/doc/stack-of-tasks/tsid/master/doxygen-html/ ; but I don't know how to do this yet.

Anyway, this is an useful resource, so I think we could merge this as it, but I'll give some time for others to react.

Issue #76, opened 2020-03-30
Use https://github.com/stack-of-tasks/eiquadprog


nim65s:
Hi,

@gabrielebndn externalized eiquadprog code into another package, so we should use this one here.

Who wants to handle that ?

If no one is available, I can give a try.

Also, this dependency is LGPL-v3, so this would fix #20

andreadelprete:
Unfortunately I'm not able to handle this right now, but I can review the pull request if someone does it.

andreadelprete:
BTW, this was already discussed in #9

PR #85, opened 2020-05-10
Add contributors and LICENSE file

### LINKED ISSUE #20: BSD License
Dear all,
We will soon switch to BSD License v2.0 to ease our collaborations.
Let me know if there is a problem with this.
Cheers,
Olivier.

olivier-stasse:
Add contributors (maybe the name of the organization should be fixed to reflect when people made their contributions)
Fix #20

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,25 @@
+BSD 2-Clause License
+
+Copyright (c) 2019, CNRS, Univ. of Trento, INRIA, Max Planck Institure
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


nim65s:
I'm not sure this contributors file is useful. We already have:
1. a section in the README
2. metadata in package.xml
3. git history

olivier-stasse:
Thanks Guilhem. You are right. It is now removed.

nim65s:
Ok, thanks. I'll mark this PR as a draft, to avoid its merge, as we should fix #76 first.
Then, we should not merge it without an history rewrite, as there is a commit that creates a file, and another one that deletes it. A squash should be enough.

Issue #88, opened 2020-05-11
Force discontinuities when adding contact


pFernbach:
Hello,
When adding a new contact, the contact forces of the other contacts computed by the `InverseDynamicsFormulationAccForce` solver are discontinuous. See the plot below that shows the forces along z of the 4 feet of a quadrupedal. At the beginning, only 3 feet are in contact, at t=6.10 the LF foot create contact. 

![contact_normal_force](https://user-images.githubusercontent.com/11633215/81592973-11822580-93bf-11ea-9772-351bb2bb08a0.png)

As you can see, the total force is immediately (in one iteration) distributed between the 4 contacts, which create discontinuities. 

Is there any way in the current implementation to have a 'smooth' change of the force distribution after adding a new contact or do I have to compute a force reference myself and send it with `contact.setForceReference()` ?

andreadelprete:
I think there're two ways to achieve this.

### Adjust the regularization weights
When you [add a new contact](https://github.com/stack-of-tasks/tsid/blob/master/src/formulations/inverse-dynamics-formulation-acc-force.cpp#L230) you can specify the regularization weight. If you initially set this weight quite high (e.g., 1) with a zero reference, I expect the initial force is gonna be almost zero. Then you can smoothly [change this weight](https://github.com/stack-of-tasks/tsid/blob/master/src/formulations/inverse-dynamics-formulation-acc-force.cpp#L265) to reach a small value (e.g., 1e-5), which will allow the contact force to increase.

### Use a time-varying upper bound on normal force
This is already done when you [remove contacts](https://github.com/stack-of-tasks/tsid/blob/master/src/formulations/inverse-dynamics-formulation-acc-force.cpp#L547) so it shouldn't be too hard to do the same when you add contacts, but it's still more work than the previous solution because it requires you to modify the code. The idea is to have a time-varying bound on the normal force, so that it's forced to change smoothly. For the case when you remove a force, what's currently implemented is an upper bound that goes from the current value to almost zero as a linear function of time. The user can decide how fast this transition is using the parameter `transition_duration`.

pFernbach:
This is the result with your first solution:

![contact_normal_force-1](https://user-images.githubusercontent.com/11633215/81797783-84021b00-950f-11ea-979a-7abf1c87e162.png)

It is working as expected, thanks. I had to add `updateRigidContactWeights` to the python API, I will make a PR.

andreadelprete:
Nice!

PR #91, opened 2020-05-13
updateRigidContactWeights: fix and add to python API

pFernbach:
Related to https://github.com/stack-of-tasks/tsid/issues/88

### DIFF
--- a/src/formulations/inverse-dynamics-formulation-acc-force.cpp
+++ b/src/formulations/inverse-dynamics-formulation-acc-force.cpp
@@ -270,7 +270,7 @@ bool InverseDynamicsFormulationAccForce::updateRigidContactWeights(const std::st
     ConstraintLevel::iterator itt;
     bool force_reg_task_found = false;
     bool motion_task_found = false;
-    for(unsigned int i=1; i<m_hqpData.size(); i++)
+    for(unsigned int i=0; i<m_hqpData.size(); i++)
     {
       for(itt=m_hqpData[i].begin(); itt!=m_hqpData[i].end(); itt++)
       {
--- a/src/formulations/inverse-dynamics-formulation-acc-force.cpp
+++ b/src/formulations/inverse-dynamics-formulation-acc-force.cpp
@@ -270,7 +270,7 @@ bool InverseDynamicsFormulationAccForce::updateRigidContactWeights(const std::st
     ConstraintLevel::iterator itt;
     bool force_reg_task_found = false;
     bool motion_task_found = false;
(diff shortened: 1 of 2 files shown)


src/formulations/inverse-dynamics-formulation-acc-force.cpp
```diff
@@ -270,7 +270,7 @@ bool InverseDynamicsFormulationAccForce::updateRigidContactWeights(const std::st
     ConstraintLevel::iterator itt;
     bool force_reg_task_found = false;
     bool motion_task_found = false;
-    for(unsigned int i=1; i<m_hqpData.size(); i++)
+    for(unsigned int i=0; i<m_hqpData.size(); i++)
```
andreadelprete, line 273: I think the reason for skipping tasks at level zero is that they are just constraints, so changing the weights does not have any effect on them. Would you agree?

pFernbach: I see, but by default the "motion_task" of the contact is at level zero and this test always fail: https://github.com/stack-of-tasks/tsid/blob/devel/src/formulations/inverse-dynamics-formulation-acc-force.cpp#L285

pFernbach: I propose to revert this change of index but invert the 
`if(motion_weight>=0.0)` and `if(itt->second->name() == contact_name+"_motion_task")`.
That way it return False only if the user requested a change in the `motion_weight` while `motion_task` was in the level 0.

And the method can return True as expected if only a change of `force_regularization_weight` is requested.

andreadelprete: Yes, if the motion task is at level zero then its weight is not gonna be changed because changing it would not have any effect on the problem solution. However, one could also decide to have motion tasks at level 1, and in that case the weight would be changed. I don't see any problem with this, but maybe I am missing something.

pFernbach: The problem is that in this case the method `updateRigidContactWeights` return False, which should the indication of a failure to apply the changes (or that something went wrong) while I was successfully executed.

pFernbach: I reverted the changes of index and just changed the tests to return True in this case (`motion_task` in level 0 and no change of `motion_weight` requested).
Tell me if this change is OK for you.

pFernbach:
I added a comment. About the negative value, it's the default value when the method is called with only 2 arguments.

andreadelprete:
Ok, makes sense!

PR #100, opened 2020-07-11
[README.md] Fix informations (License + pipeline)

olivier-stasse:
Congrats for the summer school on Memmo that was very nice.
This is a minor PR.
I am proposing to fix some informations such as updating the contributions and propose a paper to cite.

Maybe making a link towards the summer school notes and slides would be nice.

Finally: would you consider to put the wiki on doxygen ?
It could be included in the packages and pushed on the gh-pages of github.


andreadelprete:
Thanks for the PR @olivier-stasse . Adding links to the summer school material seems a good idea. I'll take care of it.
Putting the wiki on doxygen may be useful too, but I have no idea is there's an automatic way to do it. Maybe @nim65s knows?

Issue #77, opened 2020-04-01
angular-momentum-equality task is not using ref.pos


pFernbach:
The task-angular-momentum-equality take the reference momentum from the `trajectorySample.vel()` value (https://github.com/stack-of-tasks/tsid/blob/master/src/tasks/task-angular-momentum-equality.cpp#L98) and it's derivative from the `trajectorySample.acc()` value (https://github.com/stack-of-tasks/tsid/blob/master/src/tasks/task-angular-momentum-equality.cpp#L103).

I think that I understand the reasoning here that a momentum is relative to a 'velocity'. But as the name of the task suggest I would have expected to set the momentum reference from the `trajectorySample.pos()` value and it's derivative from the `trajectorySample.vel()` value. 

Can I propose a PR to change this or should we keep the current behaviour ?

andreadelprete:
I don't have a strong opinion on this, so I'd tend to agree with you, unless somebody thinks this is a good idea. I suggest to wait a couple of days to make sure nobody is against this for some good reasons, and then you can change it.

jcarpent:
I think the naming pos, vel and acc are misleading in fact. Another naming would be expected here I would say.

andreadelprete:
Maybe we could replace `vel` with `rate_of_change`? But what could we use to replace `pos`? And `acc`?

pFernbach:
It could be `value`, `time_derivative`,
 `time_second_derivative` ?

andreadelprete:
I like `value`! Maybe I'd go for `first_time_derivative` and `second_time_derivative` though.

jcarpent:
Yes, I do agree that `value` seems more appropriete.

andreadelprete:
@pFernbach Do you plan to do the PR? Or should I do it instead?

pFernbach:
Sorry I forgot about, I'll do it next week. Do you confirm the following changes:
* rename `TrajectorySample::pos, vel, acc` to `value, time_derivative, second_time_derivative`
* Adapt all the C++ code
* Set the current python API with `pos, vel, acc` deprecated and add a new one with `value, time_derivative, second_time_derivative`

andreadelprete:
Yes, that was the plan, but now that I think about it, since a trajectory is a function of time only, maybe we don't need to specify that the derivatives are wrt time. Therefore, we could simplify the names to `value, derivative, second_derivative`. What do you think?

pFernbach:
[...]

PR #101, opened 2020-07-16
Refactor TrajectorySample::pos, vel, acc to value, derivative, second_derivative

pFernbach:
Fix https://github.com/stack-of-tasks/tsid/issues/77

I did not find a good way to define "deprecated" methods with boost_python, so right now both API exist in python which is not really good. 

Feel free to add a commit in this PR or point me to a way to fix this.

### DIFF
--- a/include/tsid/trajectories/trajectory-base.hpp
+++ b/include/tsid/trajectories/trajectory-base.hpp
@@ -32,16 +32,16 @@ namespace tsid
     public:
       EIGEN_MAKE_ALIGNED_OPERATOR_NEW
       
-      math::Vector pos, vel, acc;
+      math::Vector value, derivative, second_derivative;
 
       TrajectorySample(unsigned int size=0)
       {
         resize(size, size);
       }
 
-      TrajectorySample(unsigned int size_pos, unsigned int size_vel)
+      TrajectorySample(unsigned int size_value, unsigned int size_derivative)
       {
-        resize(size_pos, size_vel);
+        resize(size_value, size_derivative);
       }
 
       void resize(unsigned int size)
@@ -49,11 +49,11 @@ namespace tsid
         resize(size, size);
       }
 
-      void resize(unsigned int size_pos, unsigned int size_vel)
+      void resize(unsigned int size_value, unsigned int size_derivative)
       {
-        pos.setZero(size_pos);
-        vel.setZero(size_vel);
-        acc.setZero(size_vel);
+        value.setZero(size_value);
+        derivative.setZero(size_derivative);
+        second_derivative.setZero(size_derivative);
       }
     };
 
--- a/include/tsid/trajectories/trajectory-base.hpp
+++ b/include/tsid/trajectories/trajectory-base.hpp
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2017 CNRS
+// Copyright (c) 2017-2021 CNRS
 //
 // This file is part of tsid
 // tsid is free software: you can redistribute it
@@ -18,7 +18,12 @@
 #ifndef __invdyn_trajectory_base_hpp__
 #define __invdyn_trajectory_base_hpp__
 
+#include <pinocchio/spatial/se3.hpp>
+
+#include "tsid/deprecation.hpp"
+#include "tsid/macros.hpp"
 #include "tsid/math/fwd.hpp"
+#include "tsid/math/utils.hpp"
 
 #include <string>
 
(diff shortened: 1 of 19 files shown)


andreadelprete:
thanks for the PR @pFernbach . The only way I can think of to fix this is to make the deprecated python methods call a deprecated c++ method, which will then raise a warning message. This deprecated c++ method could simply be the method defined in the bindings, or it could be the old method (e.g., pos, vel). Do you see what I mean?

jcarpent:
In Pinocchio, there is some new feartures allowing to deprecate some Python functions. Please see there.

nim65s:
rebased on the current devel, and deprecations added

nim65s:
sot-torque-control is now correctly compiling again, while showing the deprecation warnings: everything looks good to me.

@pFernbach @andreadelprete : is this OK for you ?

include/tsid/trajectories/trajectory-base.hpp
```diff
@@ -18,43 +18,100 @@
...
+      // getters / setters with updated names for math::Vector
+      const math::Vector & getValue() const { return pos; }
+      const math::Vector & getDerivative() const { return vel; }
+      const math::Vector & getSecondDerivative() const { return acc; }
+      void setValue(const math::Vector & value) { pos = value; }
+      void setDerivative(const math::Vector & derivative) { vel = derivative; }
+      void setSecondDerivative(const math::Vector & second_derivative) { acc = second_derivative; }
+
+      // getters / setters with updated names for pinocchio::SE3
+      void getValue(pinocchio::SE3 & value) const
```
andreadelprete, line 56: If I remember correctly the initial design choices, I think this class didn't contain any specific reference to SE3 because it's the base class. Everything specific to SE3 was added in the derived class. Is there a particular reason why it's necessary to add this SE3-specific code here?

src/math/utils.cpp
```diff
@@ -37,14 +37,14 @@ namespace tsid
       vec.tail<9>() = Eigen::Map<const Vector9>(&M.rotation()(0), 9);
     }
 
-    void vectorToSE3(RefVector vec, pinocchio::SE3 & M)
+    void vectorToSE3(RefVector const vec, pinocchio::SE3 & M)
```
andreadelprete, line 40: For passing a constant ref to a vector you can use the `ConstRefVector` type.

src/tasks/task-angular-momentum-equality.cpp
```diff
@@ -96,12 +96,12 @@ namespace tsid
...
     const Vector & TaskAMEquality::momentum_ref() const
     {
-      return m_ref.vel;
+      return m_ref.getValue();
     }
 
     const Vector & TaskAMEquality::dmomentum_ref() const
     {
-      return m_ref.acc;
+      return m_ref.getSecondDerivative();
```
andreadelprete, line 104: If the momentum is the value, then the momentum derivative should be the firstDerivative, shouldn't it?

src/tasks/task-angular-momentum-equality.cpp
```diff
@@ -118,10 +118,10 @@ namespace tsid
       // Get momentum jacobian
       const Matrix6x & J_am = m_robot.momentumJacobian(data);
       m_L = J_am.bottomRows(3) * v;
-      m_L_error = m_L - m_ref.vel;
+      m_L_error = m_L - m_ref.getValue();
 
       m_dL_des = - m_Kp.cwiseProduct(m_L_error)
-                + m_ref.acc;
+                + m_ref.getSecondDerivative();
```
andreadelprete, line 124: Same here.

andreadelprete (commented):
Thanks @nim65s for completing this. There're a couple things I'd like to understand before merging (see comments below), but overall it looks good.

src/math/utils.cpp
nim65s: this file shouldn't have been added to the commit, I'll remove this

src/tasks/task-angular-momentum-equality.cpp
nim65s: This was just a no-brain replace of `acc` to `getSecondDerivative()`. This `acc` comes from the initial commit: https://github.com/stack-of-tasks/tsid/commit/3cf00526fb92dc431456bf66cea3c00d7a1ee72d#diff-59c955f0c33d9ddd73e7bd4865a99dae5c8358352cd5cd3d73305172116977c3R103

I fixed it, without any change on the tests results, so I guess this was not tested :)

include/tsid/trajectories/trajectory-base.hpp
nim65s: It's the Sample class, which is a member of the Base class. But I can try something.

andreadelprete (approved):
Thanks @nim65s . Everything is fine for me now!

andreadelprete:
Shall I merge?

nim65s:
It's OK for me :)

PR #116, opened 2020-11-18
Add a mask to task-com-equality

jbmouret:
This PR (by Pauline Maurice & Me). adds a mask to TaskComEquality. This is typically useful when you want to keep the CoM at the center of the support polygon (for instance), but you do not want to constrain the height.

### DIFF
--- a/src/tasks/task-com-equality.cpp
+++ b/src/tasks/task-com-equality.cpp
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2017 CNRS
+// Copyright (c) 2017 CNRS, Inria
 //
 // This file is part of tsid
 // tsid is free software: you can redistribute it
--- a/src/tasks/task-com-equality.cpp
+++ b/src/tasks/task-com-equality.cpp
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2017 CNRS, Inria
+// Copyright (c) 2017-2020 CNRS, Inria
 //
 // This file is part of tsid
 // tsid is free software: you can redistribute it
--- a/src/tasks/task-com-equality.cpp
+++ b/src/tasks/task-com-equality.cpp
@@ -38,12 +38,27 @@ namespace tsid
       m_p_com.setZero(3);
       m_v_com.setZero(3);
       m_a_des_vec.setZero(3);
+      m_mask.resize(3);
+      m_mask.fill(1.);
+      setMask(m_mask);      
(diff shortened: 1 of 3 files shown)


jcarpent (approved):
Nice. Thanks Pauline and Jean-Baptiste.
Maybe it would be nice to add the Inria copyright.

jbmouret:
What is your convention for copyright? a line/year for each change in the file?
```
// Copyright (c) 2017 CNRS, NYU, MPI Tübingen, Inria
```
vs 
```
// Copyright (c) 2017 CNRS, NYU, MPI Tübingen
// Copyright (c) 2020 Inria
```

jcarpent:
I don't have strict opinion on that. On my side, at least for Pinocchio ans related, I extend the current copyright on the copyright line and add the new institution name at the end.

src/tasks/task-com-equality.cpp
```diff
@@ -38,12 +38,27 @@ namespace tsid
...
+
+    void TaskComEquality::setMask(math::ConstRefVector mask)
+    {
+      assert(mask.size() == 3);
+      TaskMotion::setMask(mask);
+      int n = dim();
+      m_constraint.resize(n, m_robot.nv());
+      m_p_error_masked_vec.resize(n);
+      m_v_error_masked_vec.resize(n);
+      m_drift_masked.resize(n);
```
andreadelprete, line 55: `m_drift_masked` is a Vector3, and here you're resizing it to a value that could be smaller than 3. How can that work?

andreadelprete (changes requested):
Thanks for the PR! I think I've spotted a small mistake (see comment below), but apart from that everything looks great!

src/tasks/task-com-equality.cpp
jbmouret: Good catch! I don't know. I guess we should change it to Vector.

jcarpent: With Eigen, it can works in Release mode, and you don't have any resize.
On the other side, in Debug mode, it should assert ;)

jcarpent: But you can keep just remove this line, as soon as you have another way to store the current size mask.

jbmouret: I just copy-pasted the SE3 task... I think it's better to stay consistent.

PR #126, opened 2021-03-10
silence warnings

nim65s:
Hi,

I'm getting boring warning mails from ROS buildfarm. This should silence them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,12 @@ IF(BUILD_PYTHON_INTERFACE)
   SET(PYWRAP ${PROJECT_NAME}_pywrap)
   ADD_PROJECT_DEPENDENCY(eigenpy REQUIRED PKG_CONFIG_REQUIRES eigenpy)
   SEARCH_FOR_BOOST_PYTHON(REQUIRED)
+
+  IF(Boost_VERSION GREATER 107299)
+    # Silence a warning about a deprecated use of boost bind by boost python
+    # at least fo boost 1.73 to 1.75
+    ADD_DEFINITIONS(-DBOOST_BIND_GLOBAL_PLACEHOLDERS)
+  ENDIF()
 ENDIF(BUILD_PYTHON_INTERFACE)
 
 # Main Library
--- a/include/tsid/bindings/python/formulations/formulation.hpp
+++ b/include/tsid/bindings/python/formulations/formulation.hpp
@@ -24,6 +24,8 @@
 #include <eigenpy/eigenpy.hpp>
 #include <boost/python/suite/indexing/vector_indexing_suite.hpp>
 
+#include <pinocchio/bindings/python/utils/deprecation.hpp>
+
 #include "tsid/formulations/inverse-dynamics-formulation-acc-force.hpp"
 #include "tsid/bindings/python/solvers/HQPData.hpp"
 #include "tsid/contacts/contact-6d.hpp"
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -86,6 +86,12 @@ IF(BUILD_PYTHON_INTERFACE)
   SET(PYWRAP ${PROJECT_NAME}_pywrap)
   ADD_PROJECT_DEPENDENCY(eigenpy REQUIRED PKG_CONFIG_REQUIRES eigenpy)
   SEARCH_FOR_BOOST_PYTHON(REQUIRED)
+
+  IF(Boost_VERSION GREATER 107299)
+    # Silence a warning about a deprecated use of boost bind by boost python
+    # at least fo boost 1.73 to 1.75
+    ADD_DEFINITIONS(-DBOOST_BIND_GLOBAL_PLACEHOLDERS)
```
jcarpent, line 93: Please see how we proceed for Pinocchio.

include/tsid/bindings/python/formulations/formulation.hpp
```diff
@@ -120,7 +120,8 @@ namespace tsid
         return self.updateRigidContactWeights(contact_name, force_regularization_weight, motion_weight);
       }
       static bool addRigidContact6dDeprecated(T& self, contacts::Contact6d & contact){
-        return self.addRigidContact(contact);
+        std::cout<<"[InverseDynamicsFormulationAccForce] Method addRigidContact(ContactBase) is deprecated. You should use addRigidContact(ContactBase, double) instead.\n";
```
jcarpent, line 123: This is not the correct way for deprecated functions. You can see how we proceed with Pinocchio.

jcarpent (changes requested):
I think a stronger fix for long term support would be much more helpful.
As TSID depends on Pinocchio, it should be easy to handle.

CMakeLists.txt
nim65s: for which purpose ?
I can also find uses of `ADD_DEFINITIONS()` in pinocchio. and about boost bind warnings, I can find this https://github.com/stack-of-tasks/pinocchio/blob/9389400a6018e97ec3822796d465a4b801630a1f/src/multibody/geometry.hxx#L10 but this doesn't fix boost bind warnings raised by boost python.

nim65s: (cf. links in the commit description)

jcarpent: See here: https://github.com/stack-of-tasks/pinocchio/blob/9389400a6018e97ec3822796d465a4b801630a1f/src/multibody/geometry.hxx#L104-L106
The thing is that we should avoid using ADD_DEFINITIONS to fix some leaks related to external packages I would say.

andreadelprete:
Hi @nim65s , thanks for the fix, I was also getting the annoying warnings from the build farm. It's not clear to me if @jcarpent  and you have converged to an agreement. Let me know if you want me to merge this to devel.

CMakeLists.txt
nim65s: I agree that I would prefer to avoid ADD_DEFINITIONS. But the method you used in that file won't work here . 
eg. even if I add a `using namespace boost::placeholders;` in `include/tsid/bindings/python/constraint/constraint-equality.hpp` I get:

```
[39/80] Building CXX object bindings/python/CMakeFiles/tsid_pywrap.dir/constraint/constraint-equality.cpp.o
Dans le fichier inclus depuis /usr/include/boost/smart_ptr/detail/sp_thread_sleep.hpp:22,
                 depuis /usr/include/boost/smart_ptr/detail/yield_k.hpp:23,
                 depuis /usr/include/boost/smart_ptr/detail/spinlock_gcc_atomic.hpp:14,
                 depuis /usr/include/boost/smart_ptr/detail/spinlock.hpp:42,
                 depuis /usr/include/boost/smart_ptr/detail/spinlock_pool.hpp:25,
                 depuis /usr/include/boost/smart_ptr/shared_ptr.hpp:29,
                 depuis /usr/include/boost/shared_ptr.hpp:17,
                 depuis /usr/include/boost/python/converter/shared_ptr_to_python.hpp:12,
                 depuis /usr/include/boost/python/converter/arg_to_python.hpp:15,
                 depuis /usr/include/boost/python/call.hpp:15,
                 depuis /usr/include/boost/python/object_core.hpp:14,
                 depuis /usr/include/boost/python/args.hpp:22,
                 depuis /usr/include/boost/python.hpp:11,
                 depuis ../include/tsid/bindings/python/constraint/constraint-equality.hpp:22,
                 depuis ../bindings/python/constraint/constraint-equality.cpp:18:
/usr/include/boost/bind.hpp:36:1: note: « #pragma message: The practice of declaring the Bind placeholders (_1, _2, ...) in the global namespace is deprecated. Please use <boost/bind/bind.hpp> + using namespace boost::placeholders, or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior. »
```

The issue here is that there is an obsolete `#include <boost/bind.hpp>` inside some `/usr/include/boost/python/*.hpp`, so according to that pragma message, our only lever is this `BOOST_BIND_GLOBAL_PLACEHOLDERS`.

jcarpent: Did you try to change the inclusion order?

nim65s: Yes, even if I add the correct new `#include <boost/bind/bind.hpp>` at the very begining of `include/tsid/bindings/python/constraint/constraint-equality.hpp` (along with `using namespace boost::placeholders;`), this doesn't prevent:
- `include/tsid/bindings/python/constraint/constraint-equality.hpp` to include `boost/python.hpp` (just after)
- `boost/python.hpp` to include `boost/python/iterator.hpp`
- `boost/python/iterator.hpp` to include the old deprecated`boost/bind.hpp`
- the old deprecated`boost/bind.hpp` to include the new `boost/bind/bind.hpp` and to issue the warning if `BOOST_BIND_GLOBAL_PLACEHOLDERS` is not defined.

jcarpent: You can define locally the `BOOST_BIND_GLOBAL_PLACEHOLDERS` and undef it afterwards if needed.

PR #139, opened 2021-08-03
[CMake] set INSTALL_RPATH for python on linux

nim65s:
To avoid the need for python users to set `LD_LIBRARY_PATH` when the `CMAKE_INSTALL_PREFIX` is not a standard one.

ref. https://github.com/stack-of-tasks/eigenpy/pull/241


andreadelprete:
Thanks @nim65s !

PR #141, opened 2021-08-13
Add implementation of TaskContactForceEquality

NoelieRamuzat:
Add implementation of `TaskContactForceEquality` as a PID+ to track reference 6D forces associated to a contact.
The constraint matrix is the `ForceGeneratorMatrix` from the associated contact.
The constraint vector is defined as:
formula%20%2B%20K_D%20(\dot{f}_{des}%20-%20\dot{f}_{ext})%20%2B%20K_I%20\displaystyle\int_0^T%20(f_{des}%20-%20f_{ext})dt)

Where
- formula is the desired forces and moments
- formula is the forces and moments measured by the sensor and expressed in the same frame as  formula
- formula are the PID gains


bindings/python/tasks/task-contact-force-equality.cpp
```diff
@@ -0,0 +1,30 @@
+//
+// Copyright (c) 2021 University of Trento
```
andreadelprete, line 2: Add your institution here

include/tsid/bindings/python/tasks/task-contact-force-equality.hpp
```diff
@@ -0,0 +1,98 @@
+//
+// Copyright (c) 2021 University of Trento
```
andreadelprete, line 2: You can add your institution here

include/tsid/tasks/task-contact-force-equality.hpp
```diff
@@ -15,15 +15,87 @@
...
+      typedef math::Vector3 Vector3;
+      typedef math::ConstraintEquality ConstraintEquality;
+      typedef pinocchio::SE3 SE3;
+
+      TaskContactForceEquality(const std::string & name,
+                      		     RobotWrapper & robot,
+                               const double dt,
+                      		     const std::string & contactName="");
+
+      void setContactList(const std::vector<std::shared_ptr<ContactLevel> >  *contacts);
```
andreadelprete, line 50: I think the TaskContactForceEquality shouldn't need to know the whole list of contacts. It should just provide a specification of the task related to its own contact force, for which it just needs to know the size of its own contact force vector. Then it's the Formulation class that should take care of putting this smaller matrix in the right spot of the larger matrix that multiplies the whole force vector.

src/tasks/task-contact-force-equality.cpp
```diff
@@ -17,12 +17,122 @@
...
 
-namespace tsid
+namespace tsid {
+namespace tasks {
+
+using namespace tsid::math;
+using namespace std;
+
+TaskContactForceEquality::TaskContactForceEquality(const std::string & name, RobotWrapper & robot,
+                                                   const double dt, const std::string & contactName):
```
andreadelprete, line 29: Instead of taking the contact name you could directly take a reference to the Contact object in the constructor, which you can then you to retrieve the ForceGeneratorMatrix

src/tasks/task-contact-force-equality.cpp
```diff
@@ -17,12 +17,122 @@
...
   }
+  auto& M = m_constraint.matrix();
+  M = cl->contact.getForceGeneratorMatrix(); // 6x12 for a 6d contact
+
+  Vector forceError = m_ref.pos - m_fext.pos;
+  Vector f_ref = m_ref.pos + m_Kp.cwiseProduct(forceError) + m_Kd.cwiseProduct(m_ref.vel- m_fext.vel) 
+                 + m_Ki.cwiseProduct(m_forceIntegralError);
+  m_constraint.vector() = f_ref;
+
+  m_forceIntegralError += (forceError - 0.05 * m_forceIntegralError) * m_dt;
```
andreadelprete, line 128: That hard-coded 0.05 should be replace with a tunable parameter, so that the user can specify how fast the integrator should leak.

andreadelprete (changes requested):
Thanks for the PR. I've left a few comments about some stuff I think it'd be better to improve.

include/tsid/tasks/task-contact-force-equality.hpp
NoelieRamuzat: Yes I agree.
I think I did this because in the `inverse-dynamics-formulation-acc-force.cpp` the method used to compute the constraint for the force tasks takes the whole contacts list in argument for the CoP task:
https://github.com/stack-of-tasks/tsid/blob/096932896c4ef77f0f04a13bd8c4232b4a562151/src/formulations/inverse-dynamics-formulation-acc-force.cpp#L397

I change my method so that it only checks if the associated contact of the `TaskContactForceEquality` is in the list of contacts of the formulation problem. If it is, it uses the classical compute method; if not, it returns an error message and the empty constraint.

NoelieRamuzat:
@andreadelprete I made the requested changes and merge the devel branch in order to have the refactoring of the  `TrajectorySample`.
Tell me if there is something else to change !

include/tsid/tasks/task-contact-force-equality.hpp
andreadelprete: Ok, sounds good to me!

andreadelprete:
Thanks for the updates, I will merge the PR now!

PR #147, opened 2021-10-08
demos: fix pinocchio use

nim65s:
The romeo demo is broken since https://github.com/stack-of-tasks/pinocchio/commit/1d7de03c8c98c789001b1129038270b41426c1f2.

While here, update similar issue in the quadruped demo.


andreadelprete:
Any update on this PR @nim65s ?

PR #150, opened 2021-11-17
Task Actuation Equality

paulinejmaurice:
Fill in code for task-actuation-equality, with possibility to add a normalization vector to minimize not the raw joint torques, but the normalized torques, in order to account for differences in capacities of different joints.
By Pauline : pauline.maurice@loria.fr


src/formulations/inverse-dynamics-formulation-acc-force.cpp
```diff
@@ -116,7 +116,7 @@ bool InverseDynamicsFormulationAccForce::addMotionTask(TaskMotion & task,
 {
   assert(weight>=0.0);
   assert(transition_duration>=0.0);
-  
+
```
andreadelprete, line 119: This file should not be changed in this PR

include/tsid/tasks/task-actuation-equality.hpp
```diff
@@ -15,15 +15,56 @@
...
+                                     ConstRefVector v,
+                                     Data & data);
+
+      const ConstraintBase & getConstraint() const;
+
+      void setReference(math::ConstRefVector ref); 
+      const Vector & getReference() const;
+
+      void setNormalizationVector(math::ConstRefVector norm);
+      const Vector & getNormalizationVector() const;
```
andreadelprete, line 55: I think "normalization vector" is not clear as a name. I'd suggest calling this "weight vector".

src/tasks/task-actuation-equality.cpp
```diff
@@ -15,11 +15,100 @@
...
+      return m_constraint;
+    }
+
+    const ConstraintBase & TaskActuationEquality::compute(const double ,
+                                                        ConstRefVector ,
+                                                        ConstRefVector ,
+                                                        Data & )
+    {
+      for(unsigned int i=0; i<m_activeAxes.size(); i++)
+        m_constraint.vector()(i) = m_ref(m_activeAxes(i)) * m_norm(m_activeAxes(i));
```
andreadelprete, line 108: Since the constraint vector depends only on `ref`, `norm` and `activeAxes`, wouldn't it make more sense to compute it only when one of these 3 quantities is updated?

andreadelprete (changes requested):
Thanks for the contribution, that's really a nice addition to TSID! I've left a few comments about some things that could be slightly improved before merging.

jcarpent (commented):
@paulinejmaurice You can add the LORIA copyright on top of the files ;)
Thanks indeed for the clean contribution.

include/tsid/tasks/task-actuation-equality.hpp
paulinejmaurice: Done

src/tasks/task-actuation-equality.cpp
paulinejmaurice: Agreed and done.

src/formulations/inverse-dynamics-formulation-acc-force.cpp
paulinejmaurice: Done, I think (let me know if there is an issue here, as I am not 100% sure if I did that correctly).

andreadelprete:
Thanks for the fixes, I'm gonna merge it now!

PR #151, opened 2021-11-22
Use generated config.hh and deprecated.hh rather than the custom ones.

francois-keith:


jcarpent (approved):
@francois-keith Good to have you on board. Small question: for which purpose do you use TSID?

include/tsid/tasks/task-joint-posture.hpp
```diff
@@ -21,7 +21,7 @@
 #include <tsid/tasks/task-motion.hpp>
 #include <tsid/trajectories/trajectory-base.hpp>
 #include <tsid/math/constraint-equality.hpp>
-#include <tsid/deprecation.hpp>
+#include <tsid/deprecated.hh>
```
andreadelprete, line 24: Why this change of name? Is it just a convention or is there some good reason?

jcarpent: These files are automatically generated by the cmake module.
The defaults extensions is .hh. You can personalize them using a specific macro.

francois-keith: I think this is the convention. The file deprecated.hh is the one generated by the cmake submodule (as well as config.hh), this will make the updates on those file easier to backport.

andreadelprete: Got it, thanks for the explanation!

Issue #152, opened 2021-11-23
Choose floating-base (or not) when creating a robot wrapper


EtienneAr:
Hello,

I wanted to create a robot from a pinocchio model using the constructor `RobotWrapper(const Model & m, bool verbose=false)` (or more precisely its Python binding).

However, the robot is considered to have a floating base by default and mine hasn't. I didn't find a way to change it using the current code, so I wrote those 2 lines of code : 7d795cf4a7293f25e0df4bd523b203082c104b51 .

Should I PR this change, or was there already a way to do it ?
If I PR, is this the good way to give the user a choice ?
Should I also update for the 2 other constructors (and bindings)?

andreadelprete:
Dear @EtienneAr , thanks for your message. There's already a way to use fixed-base robots with TSID:

https://github.com/stack-of-tasks/tsid/blob/master/exercizes/ex_0_ur5_joint_space_control.py#L25

You just need to set the third (optional) input argument of the constructor of RobotWrapper to false, as in the example above.

EtienneAr:
Yes, but this works only if you create a robot from a file.

In my case, I would like to create a robot wrapper from a pinocchio model (it doesn't rely on any urdf file).

There is only one constructor in this case, and it assumes a floating base : https://github.com/stack-of-tasks/tsid/blob/b0d6bff80292fb3451ca7ca438a4ab84b5b8c022/src/robots/robot-wrapper.cpp#L47

andreadelprete:
Ok, now I see. In this case you can do a PR to merge your contribution in TSID. However, I would suggest following the same approach used for the other constructor, that is to take as input parameter a `const pinocchio::JointModelVariant & rootJoint` (rather than a `bool`).

EtienneAr:
Closing since #154 has been merged.
Thank you.

PR #154, opened 2021-11-23
Choose floating-base (or not) when creating a robot wrapper

EtienneAr:
(Following this discussion #152)

This PR replicates the behavior of `RobotWrapper::RobotWrapper(const std::string & filename, const std::vector<std::string> & , bool verbose)` and  `RobotWrapper::RobotWrapper(const std::string & filename, const std::vector<std::string> & , const pinocchio::JointModelVariant & rootJoint, bool verbose)`, in terms of root joint management, for the constructors based on a `pinocchio::Model`.

⚠️ **Warning** : This PR breaks backward compatibility, as now the basic constructor from a `pinocchio::Model` creates a robot without a FreeFlyer root joint...

If we want to ensure backward compatibility using this approach, then we will create sets of constructors that will have opposite default behavior. Thus it can be very confusing and source of error...

What are your opinions on this ?

### DIFF
--- a/src/robots/robot-wrapper.cpp
+++ b/src/robots/robot-wrapper.cpp
@@ -60,7 +60,6 @@ namespace tsid
                                bool verbose)
       : m_verbose(verbose)
     {
-      std::cout<<"[RobotWrapper] Constructor RobotWrapper(const pinocchio::Model&, bool) is deprecated. You should use RobotWrapper(const pinocchio::Model&, const pinocchio::JointModelVariant&, bool) instead.\n";
       m_model = m;
       m_model_filename = "";
       m_na = m_model.nv-6;
--- a/src/robots/robot-wrapper.cpp
+++ b/src/robots/robot-wrapper.cpp
@@ -60,20 +60,29 @@ namespace tsid
                                bool verbose)
       : m_verbose(verbose)
     {
+      std::cout<<"[RobotWrapper] Constructor RobotWrapper(const pinocchio::Model&, bool) is deprecated. You should use RobotWrapper(const pinocchio::Model&, const pinocchio::JointModelVariant&, bool) instead.\n";
       m_model = m;
       m_model_filename = "";
-      m_na = m_model.nv;
+      m_na = m_model.nv-6;
       init();
     }
 
     RobotWrapper::RobotWrapper(const pinocchio::Model& m,
-                               const pinocchio::JointModelVariant & rootJoint,
+                               RootJointType rootJoint,
                                bool verbose)
       : m_verbose(verbose)
     {
       m_model = m;
       m_model_filename = "";
-      m_na = m_model.nv-6;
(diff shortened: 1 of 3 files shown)


andreadelprete:
Good point about the breaking of backward compatibility. I think the constructor based on a pinocchio model is much less used than the one based on a URDF file, but it's worth considering a way to make this transition as smooth as possible for everybody. Maybe in the new constructor taking a pinocchio model we could now have the `rootJoint` as a mandatory parameter. In this way, if somebody was using that constructor, they would at least immediately see that there was a change in the API, and would quickly identify and fix the problem. This seems better than the subtle error they would get if we don't do this. What do you think?

EtienneAr:
This seems like a good idea. But if `rootJoint` is mandatory, how to create a robot with a fixed base ?

andreadelprete:
> This seems like a good idea. But if `rootJoint` is mandatory, how to create a robot with a fixed base ?
> 
> https://github.com/stack-of-tasks/pinocchio/blob/593d4d43fded997bb9aa2421f4e55294dbd233c4/src/multibody/joint/joint-collection.hpp#L69-L82

I was expecting pinocchio to have a fixedJoint type, which could be used for fixed base robots, but indeed there seems to be none. Maybe @jcarpent could help us here...

jcarpent:
You should have two constructors, one with root_joint, the other one without. This how we do for the URDF parser for instance.

andreadelprete:
> You should have two constructors, one with root_joint, the other one without. This how we do for the URDF parser for instance.

That was our initial idea as well. However, that breaks backward compatibility because at the moment there's already a constructor without the `root_joint` input parameters, but this constructor assumes there's a floating base. Therefore, this is gonna break all the code that uses this constructor, resulting in a subtle error because the number of DoFs of the robot model is not gonna be different from what people expected.

jcarpent:
I would say the best way would be to add a Enum FLOATING_BASE_SYSTEM, FIXE_BASE_SYSTEM and add as input to the constructor without root_joint.
It will be nice to also deprecate the default constructor, which was certainly too restrictive.

src/robots/robot-wrapper.cpp
```diff
@@ -65,6 +56,36 @@ namespace tsid
       init();
     }
 
+    RobotWrapper::RobotWrapper(const pinocchio::Model& m,
+                               bool verbose)
+      : m_verbose(verbose)
+    {
+      std::cout<<"[RobotWrapper] Constructor RobotWrapper(const pinocchio::Model&, bool) is deprecated. You should use RobotWrapper(const pinocchio::Model&, const pinocchio::JointModelVariant&, bool) instead.\n";
```
andreadelprete, line 63: Please remove this print

andreadelprete (approved):
The code seems good to me, there's just a cout to remove.

Issue #160, opened 2022-03-04
Joint posture task computation incorrect for robot with `nq != nv`


EtienneAr:
Hello,

I noticed a computation error for the joint posture task if `nq != nv` : 
https://github.com/stack-of-tasks/tsid/blob/b0d6bff80292fb3451ca7ca438a4ab84b5b8c022/src/tasks/task-joint-posture.cpp#L157-L163

The configuration vector is cropped to the wrong size to remove the root joint (because `nq != nv`):
https://github.com/stack-of-tasks/tsid/blob/b0d6bff80292fb3451ca7ca438a4ab84b5b8c022/src/tasks/task-joint-posture.cpp#L157

And the error is compute with a term by term difference:
https://github.com/stack-of-tasks/tsid/blob/b0d6bff80292fb3451ca7ca438a4ab84b5b8c022/src/tasks/task-joint-posture.cpp#L159

Instead I propose this implementation (using `pinocchio::difference` and by cropping the error, instead of the input configuration)
```
m_v = v.tail(m_robot.na()); 

m_p_error = pinocchio::difference(m_robot.model(), m_p, m_ref.getValue());
m_p_error = m_p_error.tail(m_robot.na());

m_v_error = m_v - m_ref.getDerivative(); 
m_a_des = - m_Kp.cwiseProduct(m_p_error) 
          - m_Kd.cwiseProduct(m_v_error) 
          + m_ref.getSecondDerivative(); 
```

I will open a PR, because this will cause some other small changes in the code.

Also, I think this mistake is done at several places in the code. I could try to track them and patch them too.
What do you think ?

jcarpent:
Very nice initiative @EtienneAr. This is fine on my side.

olivier-stasse:
Thanks @EtienneAr and @jcarpent for the comment and the review.
I am a bit surprise because we have been using this task quite a lot.
Where does the problem comes from ?
https://github.com/stack-of-tasks/tsid/blob/b0d6bff80292fb3451ca7ca438a4ab84b5b8c022/src/tasks/task-joint-posture.cpp#L157-L159
looks good to me.
Is this because the reference value `m_ref` is not given properly (i.e. with the free-flyer instead of without the free-flyer) ?

andreadelprete:
I think @olivier-stasse is right. There is no bug in the code. The reference q should contain only the joint angles because the posture task controls only the joint angles, without the floating base.

EtienneAr:
[...]

PR #161, opened 2022-03-04
Fix joint task computation (for robot with `nq != nv`)

EtienneAr:
Hello,

Following the issue #160, here is my proposed fix for this problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *~
 *.pyc
 Xcode/
+.vscode/
 build*/
 _build*/
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *~
 *.pyc
 Xcode/
+.vscode/
 build*/
 _build*/
 .vscode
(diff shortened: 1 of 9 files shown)


EtienneAr:
Following https://github.com/stack-of-tasks/tsid/issues/160#issuecomment-1061727441 :
I wrote a [test](https://github.com/stack-of-tasks/tsid/blob/5b1070ebf831793973903578ce23b435dc82404b/tests/python/test_Tasks.py#L260-L324) checking a posture task for a robot with `nq != nv` .

For that I added a new [robot](https://github.com/stack-of-tasks/tsid/blob/5b1070ebf831793973903578ce23b435dc82404b/models/arm/generator.py) in `/models` (it is a 7 DoFs arm with spherical joints for shoulder and wrist and a revolute joint for the elbow). The robot is created with a `.py` file, as URDF does not support spherical joints...

andreadelprete (approved):

PR #162, opened 2022-03-24
optional support of qpmad solver

ivanbergonzani:
As suggested in #2, this PR add an optional support to the qpmad solver.

Here the results from the hqp_solver test:
```
Gonna perform 1000 tests with 60 variables, 36 equalities, 40 inequalities
### TEST FINISHED ###
*** PROFILING RESULTS [ms]                                  min        avg        max        lastTime   nSamples   totalTime  ***
Eiquadprog                                                  0.352      0.397      0.878      0.431      1000       397.496   
Eiquadprog Fast                                             0.324      0.362      1.034      0.389      1000       361.694   
Eiquadprog Real Time                                        0.236      0.269      0.560      0.290      1000       269.093   
QPMAD                                                       0.226      0.254      0.547      0.270      1000       253.643   
*** STATISTICS (min - avg - max - last - nSamples - total) ***
active inequalities                                         13.0        16.9    21.0    18.0    1000    16901.0
solver iterations                                           14.0        19.7    26.0    23.0    1000    19725.0
*** No errors detected
```


jbmouret:
Another remark: we tried OSQP and it was much slower than eiquadprog in our tests, even with warm-start.

jbmouret:
Results on my macbook pro:
```
Gonna perform 1000 tests with 60 variables, 36 equalities, 40 inequalities

### TEST FINISHED ###

*** PROFILING RESULTS [ms]                                  min        avg        max        lastTime   nSamples   totalTime  ***
Eiquadprog                                                  0.262      0.316      0.659      0.330      1000       316.090   
Eiquadprog Fast                                             0.240      0.285      0.538      0.300      1000       284.865   
Eiquadprog Real Time                                        0.232      0.277      0.523      0.290      1000       276.782   
QPMAD                                                       0.222      0.261      0.521      0.272      1000       261.245   

*** STATISTICS (min - avg - max - last - nSamples - total) ***
active inequalities                                         13.0	19.8	23.0	20.0	1000	19775.0
solver iterations                                           14.0	23.9	30.0	24.0	1000	23875.0
```

jcarpent:
> Another remark: we tried OSQP and it was much slower than eiquadprog in our tests, even with warm-start.

This is expected as OSQP is a sparse solver while the TSID problems are mostly dense.

jcarpent:
> Results on my macbook pro (with clang):
> 
> ```
> Gonna perform 1000 tests with 60 variables, 36 equalities, 40 inequalities
> 
> ### TEST FINISHED ###
> 
> *** PROFILING RESULTS [ms]                                  min        avg        max        lastTime   nSamples   totalTime  ***
> Eiquadprog                                                  0.262      0.316      0.659      0.330      1000       316.090   
> Eiquadprog Fast                                             0.240      0.285      0.538      0.300      1000       284.865   
> Eiquadprog Real Time                                        0.232      0.277      0.523      0.290      1000       276.782   
> QPMAD                                                       0.222      0.261      0.521      0.272      1000       261.245   
> 
> *** STATISTICS (min - avg - max - last - nSamples - total) ***
> active inequalities                                         13.0	19.8	23.0	20.0	1000	19775.0
> solver iterations                                           14.0	23.9	30.0	24.0	1000	23875.0
> ```

The timings are quite similar. Both solutions indeed rely on the same original algorithms with slightly different implementations. We have developed at Inria Paris a new solver that we will add to TSID shortly too.
I think it is an overall nice thing to be able to switch from one solver to another one to have a clear idea of what happens.

include/tsid/solvers/solver-HQP-qpmad.hpp
```diff
@@ -0,0 +1,84 @@
+//
+// Copyright (c) 2017 CNRS
```
andreadelprete, line 2: Fix copyright here

include/tsid/solvers/solver-HQP-qpmad.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <tsid/solvers/solver-HQP-base.hpp>
+
+#include <qpmad/solver.h>
+
+namespace tsid
+{
+  namespace solvers
+  {
+    /**
+     * @brief Abstract interface for a Quadratic Program (HQP) solver.
```
andreadelprete, line 30: Fix this comment

include/tsid/solvers/solver-HQP-qpmad.hpp
```diff
@@ -0,0 +1,84 @@
...
+      Vector m_g;   // gradient vector
+      Vector m_lb;  // gradient vector
+      Vector m_ub;  // gradient vector
+      Matrix m_C;   // constraint matrix
+      Vector m_cl;  // constraints lower bound
+      Vector m_cu;  // constraints upper bound
+
+      double m_hessian_regularization;
+
+      unsigned int m_nc;  /// number of equaloty-inequality constraints
```
andreadelprete, line 78: Fix typo

src/solvers/solver-HQP-qpmad.cpp
```diff
@@ -0,0 +1,224 @@
+//
+// Copyright (c) 2017 CNRS
```
andreadelprete, line 2: Fix copyright

andreadelprete (changes requested):
Thanks for the nice contribution. I've pointed out a few minor things to be fixed before merging.

jbmouret:
> I think the factory could be useful, and if somebody does not want to use it they free to create a solver with new, so I don't see a valid reason for removing it.

I think this is a minor debate ("more like a remark"), so I do not want to push :)

However, you did not answer the question "in which case is it useful?". On the other hand, it can be interesting to remove things to simplify the library (e.g., when you are teaching or starting to use a new lib, you want to minimize the complexity of the library: do you prefer to teach using a factory? do you want to answer questions about this?).

Again, it's very minor, so no worries :)

jbmouret:
> The timings are quite similar. Both solutions indeed rely on the same original algorithms with slightly different implementations. We have developed at Inria Paris a new solver that we will add to TSID shortly too. I think it is an overall nice thing to be able to switch from one solver to another one to have a clear idea of what happens.

We are looking forward to your new solver! 

Regarding qpmad, the timings are similar but qpmad is consistently faster by 10-20%. Once we will have validated it on our robot, I think we will use it by default. It's surprising that the difference is smaller with clang than gcc (like, clang is better at optimizing eiquadprog).

andreadelprete:
> However, you did not answer the question "in which case is it useful?". 

You're right. I didn't answer because you had already explained in which case it could be useful: when you want to map a string in a config file to a solver, or similar situations (e.g., a signal inside dynamic graph).

CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ CHECK_MINIMAL_CXX_STANDARD(11 ENFORCE)
 # Project dependencies
 ADD_PROJECT_DEPENDENCY(pinocchio 2.3.1 REQUIRED)
 ADD_PROJECT_DEPENDENCY(eiquadprog 1.1.3 REQUIRED)
+FIND_PACKAGE(qpmad QUIET)
```
nim65s, line 77:
```suggestion
ADD_PROJECT_DEPENDENCY(qpmad QUIET)
```
Because the following `TARGET_INCLUDE_DIRECTORIES` is declared as `PUBLIC` and the included (and installed)  header leaks the dependency to qpmad

nim65s (changes requested):
This looks really nice, and I was able to test this without any issue. Thanks for this work !

CMakeLists.txt
ivanbergonzani: Using cmake 3.22

in case qpmad is not found, ```ADD_PROJECT_DEPENDENCY``` leads to a cmake find_package error in libraries that links to tsid.

the [find_dependency](https://cmake.org/cmake/help/v3.13/module/CMakeFindDependencyMacro.html?highlight=find_dependency#command:find_dependency) command inside the tsid cmake config file will fail, even with the ```QUIET``` keyboard.

My understanding is that ```find_dependency``` should not be used for optional components, and qpmad is supposed to be optional.

With find_package this problem does not occur. The ```TARGET_INCLUDE_DIRECTORIES``` is already computed only in case qpmad is found and installed in the system.

nim65s: You are right, I didn't think about this case. 

But if we don't want to leak the dependency, I think we should use a `PRIVATE` option for the `TARGET_INCLUDE_DIRECTORIES`, and maybe avoid installing publicly this header, which include a qpmad header.

Otherwise, a workaround would be to keep this `find_package`,and add a `ADD_PROJECT_DEPENDENCY` inside a ` if(qpmad_FOUND)` block. 

What do you think ?

ivanbergonzani: Using the ```PRIVATE``` option will force anyone that want to use this solver with tsid to ```find_package``` again manually to be able to find the correct headers.

I prefer the second way.

The header are installed only in case qpmad is found.

nim65s (approved):

andreadelprete (approved):

PR #177, opened 2022-11-10
Add support for proxqp and osqp solver

fabinsch:
Hello, with this PR I would like to propose a way to integrate **PROXQP** and **OSQP** solver into TSID. I added options to the CMakeLists file to support the solvers, which will first check if the [proxsuite](https://github.com/simple-robotics/proxsuite) and [osqp-eigen](https://github.com/robotology/osqp-eigen) are available (both can be easily installed via conda for example).

The integration is done in c++ and I added the corresponding python bindings. The unittests, both cpp and python, got modified to test also the new solvers.

In addition to the `HQPData` (holding the constraints and cost as TSID objects), this PR adds a struct called QPData which is gathering all necessary eigen matrices to describe a QP problem. Currently, these matrices are members of each solver.

I am looking forward to your feedback.

### DIFF
--- a/src/solvers/solver-HQP-eiquadprog-fast.cpp
+++ b/src/solvers/solver-HQP-eiquadprog-fast.cpp
@@ -55,24 +55,25 @@ namespace tsid
 #ifndef NDEBUG
         sendMsg("Resizing equality constraints from "+toString(m_neq)+" to "+toString(neq));
 #endif
-        m_CE.resize(neq, n);
-        m_ce0.resize(neq);
+        m_qpData.CE.resize(neq, n);
+        m_qpData.ce0.resize(neq);
       }
       if(resizeIn)
       {
 #ifndef NDEBUG
         sendMsg("Resizing inequality constraints from "+toString(m_nin)+" to "+toString(nin));
 #endif
-        m_CI.resize(2*nin, n);
-        m_ci0.resize(2*nin);
+        m_qpData.CI.resize(2*nin, n);
+        m_qpData.ci0.resize(2*nin);
       }
       if(resizeVar)
       {
 #ifndef NDEBUG
         sendMsg("Resizing Hessian from "+toString(m_n)+" to "+toString(n));
 #endif
-        m_H.resize(n, n);
-        m_g.resize(n);
+        m_qpData.H.resize(n, n);
+        m_qpData.g.resize(n);
+        m_output.x.resize(n);
       }
       
       if(resizeVar || resizeIn || resizeEq)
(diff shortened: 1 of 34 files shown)


nim65s:
This looks very good, thank you.

Moving eiquadprog to HQPData seems to be the right thing to do, but the current version is a breaking change.

Maybe we should replace them by deprecated properties, instead of removing them ?

fabinsch:
Hi @nim65s, thanks for your review. Ok, so you are saying instead of removing the protected members `m_H`, `m_g`, `m_CI` etc to a new struct, I should leave them where they are and use `TSID_DEPRECATED` in their declaration?

I thought that I could move them as they were protected members of each solver and only used inside the solvers. I took care to not break the API for the user by not changing the function signatures. 

But you are right, a problem could be if someone implemented his own solver that is inheriting from `SolverHQuadProg` for example as the members `m_H`, `m_g` etc. disappear through this PR.

jcarpent (approved):
Thanks a lot @fabinsch for the nice work.
A small mention of the fact that both OSQP and ProxQP do not require any regularization.

fabinsch:
> Thanks a lot @fabinsch for the nice work. A small mention of the fact that both OSQP and ProxQP do not require any regularization.

Thanks for the clarification. I've not seen a hessian regularization before in proxqp or osqp that's why I set the `bool hessianRegularization` to `false` by default for proxqp and osqp. If you want and think its more consistent, I can remove it entirely from both solvers.

fabinsch:
Hello @andreadelprete, I was wondering if you as the main developer of this repo might have some time to review this PR in the near future by any chance ? Thanks a lot

andreadelprete:
> Hello @andreadelprete, I was wondering if you as the main developer of this repo might have some time to review this PR in the near future by any chance ? Thanks a lot

Yes, I was following this development but could not find the time to review it yet, very bad week at home. Hopefully tomorrow I'll do it.

src/solvers/solver-HQP-eiquadprog-fast.cpp
```diff
@@ -86,105 +87,112 @@ namespace tsid
...
+              m_qpData.g.noalias() -= w*constr->matrix().transpose()*constr->vector();
+            }
+            
+            if (hessianRegularization)
+            {
+              double m_hessian_regularization(DEFAULT_HESSIAN_REGULARIZATION);
+              m_qpData.H.diagonal().array() += m_hessian_regularization;
+            }
+          }
+    }
```
andreadelprete, line 179: There seems to be an indentation issue here, which makes the changes of the PR hard to understand

andreadelprete (commented):
This is a really nice PR, with lots of improvements here and there. The only thing I am failing to understand is the benefit of introducing the new data structure QPData. Is it motivated by your need for accessing this data from outside the solvers? Apart from this there were a few small issues, for which I've left specific comments.

src/solvers/solver-HQP-eiquadprog-fast.cpp
fabinsch: Oups sorry, I corrected it.

fabinsch:
Hello @andreadelprete and thanks for your review. I integrated your comments into the PR. 

And yes, the QPData structure was meant to gather the complete package of matrices describing a general qp problem, which can then easily be exposed in python, extracted, and used in different contexts outside the solvers. What do you think about it?

nim65s:
for my part, I think this is a nice feature, but also a breaking change. Maybe we could de-couple this new structure from the addition of those solvers ? I'm pretty sure we are going to have troubles when our users will upgrade TSID and their code will stop working.

jcarpent:
I think it is hard to decouple both.
I would say we can increase the major release version.
We have also in mind to support SOCP problems in addition to QPs, meaning that will depict Coulomb friction cone constraints as a second-order cone.

fabinsch:
hey @nim65s, thanks for sharing your opinion. I am wondering what in this PR is for you the breaking change. I did not change the signature of the functions of the eiquadprog(-fast) solvers, the members are still part of the solver (marked with deprecated) but additionally, we have the new QPData struct, and I added two functions. I think it should still work for the users, but maybe I am missing something. Thanks for your help

andreadelprete (approved):
This seems ok to me now. Let's see what @nim65s says.

nim65s:
OK for me if we bump the major version :)
I'll take the opportunity to setup the code format & linters before that

jcarpent:
> OK for me if we bump the major version :) I'll take the opportunity to setup the code format & linters before that

As mentioned by @fabinsch, it seems there is no break of API in the current PR.
@nim65s Could you confirm?

jcarpent:
We have a break of API indeed, as mentioned by @nim65s in private messages.
So, let's merge this PR and then remove useless and deprecated quantities.

PR #180, opened 2022-11-19
Enhance Python target packaging

jcarpent:
Importantly, this is required if we want to make the targets compliant with PyPy.


nim65s (approved):
I don't understand why pypy need this, but this is not breaking anything for me, so we can merge :)
Maybe we could add this to jrl-cmakemodule ?

jcarpent:
I think pypy is looking for the suffix extension. ;)

PR #183, opened 2022-11-19
Clean and update contributors list

jcarpent:


andreadelprete:
This looks fine to me!

jcarpent:
@fabinsch Could you check why the CI on conda is falling?
osqp-eigen has been updated to the latest release version.

fabinsch:
Hey, so the python unittest `test_Solvers.py` was broken and deactivated in the `CMakeLists.txt` before the PR #177 where I put it back after doing this [commit](https://github.com/stack-of-tasks/tsid/pull/177/commits/f1048a422fd3fa3099fd8a35062d04cafbe9a4a4) where I fixed the `setMatrix` and `setVector` methods. Locally on my machine it is working, I just checked it, but on the CI we seems to have a problem with buffer overflow.. we can see in the CI the message `*** stack smashing detected ***: terminated` .

fabinsch:
I think the difference might be that locally I am using clang12++ while the CI is using x86_64-conda-linux-gnu-c++, which might have a different memory protection mechanism as explained [here](https://stackoverflow.com/questions/1345670/stack-smashing-detected).

PR #191, opened 2022-12-22
Input argument checking + fix virtual destructor issue

wxmerkt:
This PR adds explicit checking of input argument sizes etc. as in Pinocchio using the Pinocchio macros. Given that Pinocchio is a necessary requirement, I re-used the macros from Pinocchio rather than copy-pasting them into TSID. This is particularly important for the Python bindings which allowed wrong values to be passed in from users.

I also fixed missing virtual destructors which lead to issues when destructing TSID problems.

I've also flipped the default option for generating and installing the documentation in line with EigenPy and Pinocchio

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ set(PROJECT_DESCRIPTION
 set(PROJECT_URL "https://github.com/${PROJECT_ORG}/${PROJECT_NAME}")
 
 # Project options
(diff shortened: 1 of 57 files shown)


andreadelprete:
Thanks @wxmerkt for the work done. This looks good to me, but let's see also what @nim65s or @jcarpent think.

nim65s (approved):

wxmerkt:
Thank you for the swift review and merge @andreadelprete @nim65s. Happy holidays!

PR #193, opened 2023-01-03
Fix qpmad

wxmerkt:
`retrieveQPData` was not implemented for qpmad resulting in an abstract class preventing compilation. We aren't testing some of the optional components on CI yet so it likely got missed when refactoring when introducing ProxQP


jcarpent (approved):

jcarpent:
From a CI perspective, I do agree that tests are missing.
@wxmerkt Could you raise some missing aspects that you have noticed so far?

PR #197, opened 2023-02-08
[pre-commit] Add clang-format Google style.

olivier-stasse:
This PR add the pre-commit hook to check the C++ format. The format is following the Google style.
You need to use pre-commit to take advantage of it:
https://pre-commit.com/

Maybe this should be specified in the CONTRIBUTING.md file.

### DIFF
--- a/bindings/python/module.cpp
+++ b/bindings/python/module.cpp
@@ -30,16 +30,15 @@
 namespace bp = boost::python;
 using namespace tsid::python;
 
-BOOST_PYTHON_MODULE(tsid_pywrap)
-{
+BOOST_PYTHON_MODULE(tsid_pywrap) {
   eigenpy::enableEigenPy();
   eigenpy::exposeAngleAxis();
   eigenpy::exposeQuaternion();
 
-  typedef Eigen::Matrix<double,6,6> Matrix6d;
-  typedef Eigen::Matrix<double,6,1> Vector6d;
-  typedef Eigen::Matrix<double,6,Eigen::Dynamic> Matrix6x;
-  typedef Eigen::Matrix<double,3,Eigen::Dynamic> Matrix3x;
+  typedef Eigen::Matrix<double, 6, 6> Matrix6d;
+  typedef Eigen::Matrix<double, 6, 1> Vector6d;
+  typedef Eigen::Matrix<double, 6, Eigen::Dynamic> Matrix6x;
+  typedef Eigen::Matrix<double, 3, Eigen::Dynamic> Matrix3x;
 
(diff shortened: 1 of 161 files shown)


jcarpent (changes requested):
It should be nice to add an ignore commit to clean the diff tree.
This has been done previously in hpp-fcl and eigenpy for reference.

olivier-stasse:
@jcarpent Is this what you asked for ?

jcarpent:
Something like https://github.com/Simple-Robotics/proxsuite/blob/main/.git-blame-ignore-revs

olivier-stasse:
> Something like https://github.com/Simple-Robotics/proxsuite/blob/main/.git-blame-ignore-revs

Done. Thanks for the feedback.

bindings/python/module.cpp
```diff
@@ -15,31 +15,30 @@
...
 #include <eigenpy/geometry.hpp>
 
-#include "tsid/bindings/python/robots/expose-robots.hpp"
 #include "tsid/bindings/python/constraint/expose-constraints.hpp"
 #include "tsid/bindings/python/contacts/expose-contact.hpp"
-#include "tsid/bindings/python/trajectories/expose-trajectories.hpp"
-#include "tsid/bindings/python/tasks/expose-tasks.hpp"
-#include "tsid/bindings/python/solvers/expose-solvers.hpp"
 #include "tsid/bindings/python/formulations/expose-formulations.hpp"
+#include "tsid/bindings/python/fwd.hpp"
```
andreadelprete, line 23: Are we sure that changing the order of the include isn't gonna affect the compilation?

olivier-stasse: Good point I forgot to put the flag to avoid sorting the includes. I will do it and will check more carefully the CI tests. I got trapped in thinking that the CI was failing because of osqp (first CI failure on this PR https://github.com/stack-of-tasks/tsid/actions/runs/4126738738/jobs/7129052586).

olivier-stasse:
I think that it should be fine now. Solvers still fails so. But it is working on  my mac.

jcarpent:
I've just provided a potential fix for ROS-CI, which has already been applied to Pinocchio and EigenPy. Let's see what happens.

jcarpent:
Seems to work

jcarpent (approved):

olivier-stasse:
Thanks @jcarpent for the fix and the approval. I am waiting for @andreadelprete approval.

andreadelprete:
Looks fine to me. I'm gonna merge

PR #209, opened 2023-06-28
Fix a typo in ex_4_walking

paLeziart:
Fix a small typo in the biped walking example where the force generator matrix of the right foot was retrieved for the left foot.


andreadelprete:
Nice catch, thanks!

PR #213, opened 2023-09-16
CMake: require >= 3.10

nim65s:
ref. https://github.com/jrl-umi3218/jrl-cmakemodules/issues/338 and fix deprecation warning about CMake 3.5 from CMake 3.27

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@
 # GNU Lesser General Public License along with tsid If not, see
 # <http://www.gnu.org/licenses/>.
 
-cmake_minimum_required(VERSION 3.1)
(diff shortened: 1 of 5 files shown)


olivier-stasse:
The checks for this PR failed due to the solvers. As this was fixed with PR #216, I will merge this PR because the delay of 30 days to rerun the checks is over.
We can revert it back (or fix it) if failing.

As the modifications are very small I do not expect any problem.

olivier-stasse:
If you see any problem @andreadelprete feel free to revert it.

olivier-stasse:
https://github.com/stack-of-tasks/tsid/actions/runs/6603915478 show that it is now working.

Issue #215, opened 2023-10-19
github action on OSX conda Release failing on osqp solver


olivier-stasse:
Dear @fabinsch,
Currently unrelated PR #212 fails due to an error when testing the Oxford osqp solver on Linux /OS Conda Release.
See here for instance:
https://github.com/stack-of-tasks/tsid/actions/runs/6538220426/job/17753685229?pr=212

I suspect that this is related to an incompatibility or a release problem as the same test seems to work for Linux - ROS Noetic Debug , where as in PR #213 it was the reverse (i.e failing for Linux ROS Noetic Debug and fine for Linux/OS Conda Release).

As you authored the test, could you give an insight on how to fix the problem ?

fabinsch:
Hi @olivier-stasse,

it is correct that I added proxqp and osqp. In this test, we create random problems and then check if the solver can solve the problem up to `EPS=1e-3`.  We are not setting any accuracy for the solvers, which leads to  `epsAbs = 1e-5` for [osqp](https://github.com/stack-of-tasks/tsid/blob/master/src/solvers/solver-osqp.cpp#L35). We could try to increase the accuracy we ask for by doing `solver.set_epsilon_absolute(1e-5)` for example. Also, we can fix the random seed to be sure to always run on the same problem.

If you want, I can open another PR where I do this change, and we can check if the tests pass. If yes, we can rebase the open PRs.

olivier-stasse:
That will be awesome. Thank you very for your reactivity and your commitment.

PR #216, opened 2023-10-19
Fix solver python test

fabinsch:
In this PR, I check the returned status from the solver to make sure that the random problems were actually solved successfully. 
I fixed a random seed, like this the different CI runs are comparable. 

In the past, we had problems with osqp which did not pass the Python test, see #215. Testing on my local machine showed that osqp is reaching the maximum number of iterations. Therefore, I increased it and added the check for the status.

### DIFF
--- a/exercizes/notebooks/Installation process.ipynb
+++ b/exercizes/notebooks/Installation process.ipynb
@@ -1,5 +1,5 @@
 {
- "cells": [
+  "cells": [
   {
    "cell_type": "markdown",
    "metadata": {},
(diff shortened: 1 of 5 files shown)


jcarpent (approved):

olivier-stasse:
CI happy so I merge, because I do no think the approval from @andreadelprete is necessary here.
If this is not the case, it can be revert back.

olivier-stasse:
Thanks a lot @fabinsch for this quick PR !

Issue #165, opened 2022-07-08
TSID for "pantograph/biarticular" robot leg design?


egordv:
Hello and thank you for a great tool! It works perfectly for control of a classic rigid robot.

Right now I'm wondering what it will take to use the TSID to control a robot with an a-la biarticular leg design (something like the one mentioned [here](https://www.researchgate.net/figure/A-Pantograph-leg-with-one-spring-around-the-knee-joint-and-a-rigid-pantograph_fig1_335145762)) :
![image](https://user-images.githubusercontent.com/40291783/177970756-e286809d-b958-480f-8ffc-029ec2bf8a5e.png)

I can see three main difficulties:
1. Closed kinematic chain
2. Passive joint
3. Spring element

@andreadelprete can you please guide the right way to implement the missing features for this task (or point me that it's almost impossible in TSID framework)?

1. It looks like the pinocchio3 version is heading to support the loop constraints. Is there any plans to integrate the pinocchio3 in TSID?
2. Without pinocchio3 support, as an approximate solution of closed kinematic chain with a passive joint problem, instead of using the passive joint I can remove it and get a sort of solution by implementing a new task, something like "taskTwoFramesPositionEquality", and use it with a high gain to force the solver to keep the end of two links as close as possible, emulating the passive joint in a closed kinematic loop. Will it do the job or lead to solver failure due to over-constrained design or high-gain-induced stability problems?
3. Spring element is a mystery for me. I can model it as a prismatic/rotational joint, but what is the proper way to take into account the state-dependent spring dynamics generating the actuation in that additional joint?

andreadelprete:
Dear @egordv , thank you for the interesting questions. I'm glad you enjoy TSID.
I think it should be possible to make TSID work with this kind of robots. 

>     1. It looks like the pinocchio3 version is heading to support the loop constraints. Is there any plans to integrate the pinocchio3 in TSID?

Yes. I think we'll integrate pinocchio3 in TSID when it is officially released.
 
>     2. Without pinocchio3 support, as an approximate solution of closed kinematic chain with a passive joint problem, instead of using the passive joint I can remove it and get a sort of solution by implementing a new task, something like "taskTwoFramesPositionEquality", and use it with a high gain to force the solver to keep the end of two links as close as possible, emulating the passive joint in a closed kinematic loop. Will it do the job or lead to solver failure due to over-constrained design or high-gain-induced stability problems?

[...]

PR #218, opened 2023-10-30
Two frames contact pr

egordv:
This PR is related to the issue "TSID for "pantograph/biarticular" robot leg design" https://github.com/stack-of-tasks/tsid/issues/165  

I want to share the code I wrote to create a new type of contact following the discussion in the mentioned PR.
The main aim of this PR is to give the TSID dynamics formulation the ability to work with closed kinematic chains specified via user code (not URDF).

This PR adds two new items:
1. New motion task: [**TwoFramesEquality**](https://github.com/egordv/tsid/blob/65d780ef479e59a404ed2f14b8e8c2aed778c611/src/tasks/task-two-frames-equality.cpp#L31). This task can be used to force some frames of the robot to keep their relative displacement zero. The main difference with SE-Equality is that the TwoFramesEquality works with the relative placement of the frames, so the frames will be 'glued' to each other, not to the specified world position.
2. New contact type: [**TwoFramesContact**](https://github.com/egordv/tsid/blob/65d780ef479e59a404ed2f14b8e8c2aed778c611/src/contacts/contact-two-frames.cpp#L29). This contact uses the TwoFramesEquality motion task as it's base emulating a ball joint via using only translation part of the mask to non-zero. 

The usage of TwoFramesContact alongside with zero actuation bounds allows to create a closed kinematic chains in current TSID formulation (with current version of Pinocchio and URDF parser). The contact should be created bu user via TwoFramesContact creation.

The PR also includes a minimal example showing how to add a [closed kinematic chain](https://github.com/egordv/tsid/blob/65d780ef479e59a404ed2f14b8e8c2aed778c611/demo/demo_talos_gripper_closed_kinematic_chain.py#L88) to Talos gripper mechanism via demo/demo_talos_gripper_closed_kinematic_chain.py 

This is the sample how this gripper works without additional contact:
TalosGripperNoClosedKinematicChain

This is the sample how this gripper works **with** additional contact via TwoFramesContact thus creating a closed kinematic chain:
TalosGripperWithClosedKinematicChain

Regarding the code: there are many virtual methods in conact-base.hpp not relevant to the concept of TwoFramesContact like setContactNormal, etc. Also the underlying motion task [can be only the SE3Equality](https://github.com/stack-of-tasks/tsid/blob/f938e35d880a4f8db8655c8c9af58e356631835e/include/tsid/contacts/contact-base.hpp#L74) so the PR code contains some rudimentary methods to mitigate this in not very elegant way, may be some refactoring is required?


hrp2-14:
Hi ! This project doesn't usually accept pull requests on the main branch.
If this wasn't intentionnal, you can change the base branch of this PR to devel
(No need to close it for that). Best, a bot.

andreadelprete:
Thanks for this PR. I'm currently very busy, but I'll try to review this next week.

src/contacts/contact-two-frames.cpp
```diff
@@ -0,0 +1,227 @@
...
+  assert(frictionCoefficient>0.0);
+  if(frictionCoefficient<=0.0)
+    return false;
+  m_mu = frictionCoefficient;
+  updateForceInequalityConstraints();
+  */
+  return true;
+}
+
+bool ContactTwoFrames::setMinNormalForce(const double minNormalForce) {
```
andreadelprete, line 151: The methods for setting min and max forces should be implemented as they are used

src/contacts/contact-two-frames.cpp
andreadelprete, line 1: I would change the name of this class because the current name suggests that the whole frame is constrained, but instead it is just the position. May `ContactTwoPoints` or `ContactTwoFramePositions`.

src/tasks/task-two-frames-equality.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+const ConstraintBase& TaskTwoFramesEquality::getConstraint() const {
+  return m_constraint;
+}
+
+const ConstraintBase& TaskTwoFramesEquality::compute(const double,
+                                                     ConstRefVector,
+                                                     ConstRefVector,
+                                                     Data& data) {
+  // Calculating task with formulation: [J1 - J2   0   0] y = [-J1dot*v +
```
andreadelprete, line 119: This comment is misleading because it suggests that the constraint matrix to be computed is the one that's gonna be multiplied by the vector y containing all the decision variables. Instead, this method has to compute the matrix that multiplies just the robot accelerations `dv`.

src/tasks/task-two-frames-equality.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+  // desired acc in world frame
+  m_a_des = m_Kp.cwiseProduct(m_p_error_vec) +
+            m_Kd.cwiseProduct(m_v_error.toVector());
+
+  m_v_error_vec = m_v_error.toVector();
+
+  m_drift = (m_wMl1.act(m_drift1) - m_wMl2.act(m_drift2));
+
+  m_J1_rotated.noalias() = m_wMl1.toActionMatrix() * m_J1;
```
andreadelprete, line 155: Add comment "    // Use an explicit temporary `m_J_rotated` to avoid allocations."

src/tasks/task-two-frames-equality.cpp
```diff
@@ -0,0 +1,178 @@
...
+  m_wMl1.rotation(oMi1.rotation());
+  m_wMl2.rotation(oMi2.rotation());
+
+  m_robot.frameJacobianLocal(data, m_frame_id1, m_J1);
+  m_robot.frameJacobianLocal(data, m_frame_id2, m_J2);
+
+  // Doing all calculations in world frame
+  errorInSE3(oMi1, oMi2, m_p_error);  // pos err in local (=rotated) oMi1 frame
+  m_p_error_vec =
+      m_wMl1.toActionMatrix() * m_p_error.toVector();  // pos err in world frame
```
andreadelprete, line 142: Change comment. This vector is not in world frame, but in local-world-aligned frame, same for the quantities below

tests/python/test_Constraint.py
andreadelprete, line 1: Changes to this file are minor and should not be part of this PR.

tests/python/test_Contact.py
andreadelprete, line 1: Changes to this file are minor and should not be part of this PR.

tests/python/test_ContactPoint.py
andreadelprete, line 1: Changes to this file are minor and should not be part of this PR.

tests/python/test_Deprecations.py
andreadelprete, line 1: Changes to this file are minor and should not be part of this PR.

tests/python/test_Formulation.py
andreadelprete, line 1: Changes to this file are minor and should not be part of this PR.

tests/python/test_Gravity.py
andreadelprete, line 1: Changes to this file are minor and should not be part of this PR.

tests/python/test_Tasks.py
andreadelprete, line 1: Changes to this file are minor and should not be part of this PR.

tests/python/test_Trajectories.py
andreadelprete, line 1: Changes to this file are minor and should not be part of this PR.

andreadelprete (changes requested):
I've reviewed the code and overall it looks great. I've left some comments on small things that should be fixed before merging this PR. The main one is removing the STL/urdf files, which should not stay in this repo in my opinion. I know we have the model of a robot in TSID, but that was introduced a long time ago, and it was justified by the need to give students an easy way to use the software. Nowadays, with packages as example-robot-data, I think this is no longer necessary. Let me know your thoughts, and thanks for the great work!

include/tsid/contacts/contact-two-frame-positions.hpp
```diff
@@ -0,0 +1,132 @@
...
+   */
+  void useLocalFrame(bool local_frame);
+
+ protected:
+  void updateForceInequalityConstraints();
+  void updateForceRegularizationTask();
+  void updateForceGeneratorMatrix();
+
+  TaskTwoFramesEquality m_motionTask;
+  TaskSE3Equality m_dummyMotionTask;
```
andreadelprete, line 116: I had not noticed this in my first review, thanks for pointing it out. Indeed using this dummyMotionTask is not ideal. It would probably be better to change the API of ContactBase, so that the method `getMotionTask` returns a `MotionTaskBase` rather than an `SE3EqualityTask`. Would you agree with that?

andreadelprete (commented):
Nice job addressing all my comments, thanks! I've left just one additional comment related to the dummyMotionTask you mentioned.

include/tsid/contacts/contact-two-frame-positions.hpp
```diff
@@ -68,7 +68,7 @@ class ContactTwoFramePositions : public ContactBase {
   virtual const ConstraintEquality& computeForceRegularizationTask(
       const double t, ConstRefVector q, ConstRefVector v, const Data& data);
 
-  const TaskSE3Equality& getMotionTask() const;
+  const TaskTwoFramesEquality& getMotionTask() const;
```
andreadelprete, line 71: Shouldn't the `getMotionTask` method return a `TaskMotion`?

andreadelprete: The pure virtual method `TaskMotion& getMotionTask` of the parent class `ContactBase` should be implemented in this class, shouldn't it?

andreadelprete (commented):
Thanks for implementing these changes. There's still something unclear to me (See comments below), but apart from this, everything looks fine!

include/tsid/contacts/contact-two-frame-positions.hpp
andreadelprete: You're right. It works. I've recreated the same situation with this minimalistic code, and indeed the compiler doesn't complain. I didn't know you could do this. At this point I am fine with this PR and I am gonna merge it. Thanks again for the great contribution!

```
#include <iostream>
using namespace std;

class TaskMotion {};
class TaskTwoFramesEquality : public TaskMotion {};

class ContactBase {
    public:
    virtual const TaskMotion& getMotionTask() const = 0;
};

class ContactTwoFramePositions : public ContactBase {
    public:
    TaskTwoFramesEquality m_task;

    const TaskTwoFramesEquality& getMotionTask() const{
        cout<<"TaskTwoFramesEquality& getMotionTask"<<endl;
        return m_task;
    }
};

int main(){
    ContactTwoFramePositions c;
    const TaskMotion& task = c.getMotionTask();
    const TaskTwoFramesEquality& task2 = c.getMotionTask();
    return 0;
}
```

PR #245, opened 2024-11-20
[fix] missing const in arguments of `RobotWrapper.set_rotor_inertia` and `RobotWrapper.set_gear_ratios` python bindings

domrachev03:
This MR adds missing `const` identifier to the arguments of two methods of `RobotWrapper`: `set_rotor_inertia` and `set_gear_ratios` in python bingings. This problem practically disallowed to use motor inertia in python bindings.

Right now, this functionality does not work, and an attempt to run the simplest example:
```python
from copy import deepcopy

import numpy as np
import pinocchio as pin
import tsid
from robot_descriptions.iiwa7_description import URDF_PATH

pin_model = pin.buildModelFromUrdf(URDF_PATH)
robot = tsid.RobotWrapper(pin_model, False)

robot.set_gear_ratios(np.ones(12))
robot.set_rotor_inertias(np.ones(12))

print(robot.rotor_inertias)
print(robot.gear_ratios)
```

results in:
```bash
Boost.Python.ArgumentError: Python argument types in
    RobotWrapper.set_gear_ratios(RobotWrapper, numpy.ndarray)
did not match C++ signature:
    set_gear_ratios(tsid::robots::RobotWrapper {lvalue}, Eigen::Matrix<double, -1, 1, 0, -1, 1> {lvalue} gear ratio vector)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Fix missing `const` specifier in python bindings for methods `RobotInertia.set_rotor_inertia()`
+ and `RobotInertia.set_gear_ratios()`
 ## [1.7.1] - 2024-08-26
 
 - Fix a typo in ex_4_walking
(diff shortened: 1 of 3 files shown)


nim65s:
Thanks !
Could you add a regression test, to avoid this issue to happen again ?

domrachev03:
Done! I've modified a `tests/python/test_RobotWrapper` test to add a motor's inertias and check that it's computed correctly.

I am not 100% sure I correctly understood what do you mean by regression test, and if you want me to implement smth else, I would be glad to do that!

nim65s (approved):
You did exactly what was expected, thanks !

domrachev03:
Glad to hear that! Since the pipeline is failing in the devel as well, I guess we are good to merge those changes? Or something else is missing?

PR #251, opened 2025-03-10
Python bidings for the measured force contact

aleolive38:
Hi, 

I noticed there was no python bindings for the measured force contacts, so I implemented them following the recommendation made for a similar issue.
I added the measure6Dwrench contact type to the bindings, the addMeasuredForce binding in the formulation and a test script that checks the creation of this contact, the setting of its value and its addition to the inverse dynamics formulation.
Let me know if it is ok or if you have any suggestions, 

Have a nice day


nim65s (changes requested):
Thanks for this !

Could you add an entry in `tests/python/CMakeLists.txt` for your test ?

bindings/python/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ set(${PYWRAP}_SOURCES
     contacts/contact-6d.cpp
     contacts/contact-point.cpp
     contacts/contact-two-frame-positions.cpp
+    contacts/measured-6Dwrench
```
jcarpent, line 22: missing file extension
I would rename the file following the naming convention of the project.

include/tsid/bindings/python/contacts/measured-6Dwrench.hpp
```diff
@@ -0,0 +1,114 @@
+//
+// Copyright (c) 2022 CNRS INRIA LORIA
```
jcarpent, line 2: Sure about the year?

include/tsid/bindings/python/formulations/formulation.hpp
```diff
@@ -273,6 +276,10 @@ struct InvDynPythonVisitor
                                          const solvers::HQPOutput& sol) {
     return self.getContactForces(name, sol);
   }
+  static bool addMeasuredForce(T& self,
+                               contacts::Measured6Dwrench measured_force) {
```
jcarpent, line 280: Missing reference.

include/tsid/bindings/python/contacts/expose-contact.hpp
```diff
@@ -21,19 +21,22 @@
...
 void exposeContact6d();
 void exposeContactPoint();
 void exposeContactTwoFramePositions();
+void exposeMeasured6Dwrench();
 
 inline void exposeContact() {
   exposeContact6d();
   exposeContactPoint();
   exposeContactTwoFramePositions();
+  exposeMeasured6Dwrench();
```
andreadelprete, line 37: For consistency with the already existing "Contact6d" I would prefer using "Measured6dWrench", without capitalizing the "d" after 6.

include/tsid/bindings/python/contacts/expose-contact.hpp
```diff
@@ -21,19 +21,22 @@
 #include "tsid/bindings/python/contacts/contact-6d.hpp"
 #include "tsid/bindings/python/contacts/contact-point.hpp"
 #include "tsid/bindings/python/contacts/contact-two-frame-positions.hpp"
+#include "tsid/bindings/python/contacts/measured-6Dwrench.hpp"
```
andreadelprete, line 24: Even here, for consistency, I would rename this file "measured-6d-wrench.hpp"

andreadelprete (changes requested):

aleolive38:
Hi, 
Thanks for the reviews!
For the naming convention, I don't mind changing them though I just used the existing name from the existing cpp files.
Do you want them to be changed only in the bindings you pointed out or in the rest of the code too? Then I guess also the Measured3Dforce should be changed to Measured3dForce and the corresponding files to measured-3d-force  (I did not implement bindings for this one in this pull request though)

andreadelprete:
You are right that we already have inconsistencies. However, we should at least avoid adding more inconsistencies to the code. If you also wanna fix the naming inconsistencies of the 3d force that'd be an added value for this PR.

aleolive38:
For now I fixed the inconsistent naming only on the bindings side

aleolive38:
Hi, 
I fixed the inconsistencies also on the cpp side for the measured-3d-force and measured-6d-wrench so that all of the file names are more coherent with the rest of the project.
Let me know if this is what you had in mind.

aleolive38:
Thanks for the answer @nim65s, 

I put back the deprecated files with warnings, is this what you had in mind?

nim65s (approved):
Yes, thanks !

Issue #301, opened 2025-12-23
Adding interface to DAQP to support Hierarchical Quadratic Programming


darnstrom:
Hi @nim65s and @andreadelprete 
I am a developer of the dual active-set QP solver [DAQP](https://github.com/darnstrom/daqp). DAQP is often very efficient for dense quadratic programming problems that arise in real-time control problems. Recently, we have added support for solving hierarchical quadratic programs (see our recent [preprint](http://arxiv.org/abs/2512.18458) for details and numerical experiments). Would it be of interest to add an interface to DAQP in TSID? I see that the solvers interfaced in TSID currently only support hierarchies with two levels, but DAQP could handle multi-level hierarchies (addressing, for example #263.

If it sounds like something relevant for TSID, I would happily start working on a PR to integrate DAQP.

andreadelprete:
Hi @darnstrom , I believe it would be a nice addition to TSID. The hierarchical QP is indeed the main missing feature of this library. Moreover, the library's API has been thought to support hierarchies, so it should be relatively easy to integrate your solver in TSID. Thanks for the offer!

andreadelprete:
Hi @darnstrom is there any update on this?

darnstrom:
Thanks for the reminder @andreadelprete! I started this half a year ago, but then other projects took over. I did some wrapping up now, and  I have something that seems to be working. I want to do a bit more testing and refining (there are some overhead from rescaling since the hierarchical QP implementation of daqp works on a normalized problem.)

I am currently working on this branch [here](https://github.com/darnstrom/tsid/tree/daqp-interface), and I will hopefully open a PR here very soon.

nim65s:
Fixed in #326

PR #326, opened 2026-07-12
Daqp interface

darnstrom:
## Summary

This PR adds DAQP as an optional TSID HQP solver (see #301 for some more context.)
Specifically this PR:

- Adds DAQP solver integration, factory support, bindings, and tests.
- Supports arbitrary hierarchical QPs through DAQP's native hierarchy handling.
- Uses a conventional dense QP for the common two-level case:
    - level 0: hard constraints
    - level 1: weighted equality tasks

- Reuses the DAQP workspace across solves, including primal/dual warm-start state.
- Adds SolverHQPDAQP::setAssumeMatricesUnchanged(bool) for explicit factorization and transformed-matrix reuse when the caller knows matrices, weights, hierarchy structure, and hard bounds are unchanged. _Keeps full numerical refreshes as the default behavior_.

## Benchmark

Measured on TSID's existing 60-variable HQP solver workload with 5,100 solves per solver:

| Solver | Avg ms | p95 ms | Max ms |
| :--- | :---: | :---: | :---: |
| DAQP full refresh, warm | 0.150 | 0.163 | 0.356 |
| DAQP full refresh, cold | 0.189 | 0.214 | 0.365 |
| DAQP reuse, warm | 0.050 | 0.064 | 0.344 |
| DAQP reuse, cold | 0.089 | 0.133 | 0.334 |
| EiQuadProg | 0.222 | 0.237 | 0.639 |
| EiQuadProg Fast | 0.209 | 0.223 | 0.852 |
| EiQuadProg RT | 0.206 | 0.219 | 0.523 |
| ProxQP | 0.416 | 0.523 | 1.616 |

Reuse performs a full update after each perturbation and enables reuse only for the following unchanged solves.

## Validation

- DAQP multilevel, weighted two-level, warm-start, factor reuse, and solution-agreement checks pass.

## Installation
To build the optional DAQP solver, install [DAQP](https://github.com/darnstrom/daqp) and add `-DBUILD_WITH_DAQP=ON` to the CMake command.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,7 @@ option(
 )
 option(BUILD_WITH_PROXQP "Support using the proxqp solver" OFF)
 option(BUILD_WITH_OSQP "Support using the osqp solver" OFF)
+option(BUILD_WITH_DAQP "Support using the DAQP hierarchical QP solver" OFF)
 
 # With pos, vel, acc awaiting renaming (e.g. in trajectory-base), we are
 # producing a ton of deprecation warnings. Ignoring them for now; remove this
@@ -109,6 +110,10 @@ endif(BUILD_PYTHON_INTERFACE)
 
 add_project_dependency(pinocchio 2.3.1 REQUIRED)
 add_project_dependency(eiquadprog 1.1.3 REQUIRED)
+
+if(BUILD_WITH_DAQP)
+    add_project_dependency(daqp REQUIRED CONFIG)
+endif()
 
 find_package(qpmad QUIET) # optional
 if(qpmad_FOUND)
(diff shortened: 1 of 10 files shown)


andreadelprete (commented):
Everything looks good to me, thanks for the great work! Maybe @nim65s you also wanna take a look at it?

nim65s (approved):
Yes, I already reviewed this and I also think this is good and could be merged. There is a simple packaging point where I would like to discuss a bit, but that should be handled in daqp upstream, so probably we can already merge here and address that later :)

And ROS CI is not happy, but I'll also handle that in another PR.

Anyways, thanks a lot @darnstrom !

nim65s:
Also https://github.com/NixOS/nixpkgs/pull/542223 is waiting for review

